PR #73, opened 2015-09-14
Using preferred encoding to decode multibyte text in ListView.

belobrov-andrey:
This patch fix text encoding in ListView in multibyte applications.
locale.getpreferredencoding() method return encoding which correspond to the settings "Language for non-unicode programs" in Windows Control Panel.
It seems that there are other places where 'utf-8' endoding is hardcoded and should be fixed.

### DIFF
--- a/pywinauto/controls/common_controls.py
+++ b/pywinauto/controls/common_controls.py
@@ -27,6 +27,7 @@ from __future__ import print_function
 import time
 import ctypes
 import warnings
+import locale
 
 from .. import six
 from .. import win32functions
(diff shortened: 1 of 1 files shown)


vasily-v-ryabov:
Thank you, Andrey! We have no unit tests on multi-byte ListView yet ([see coverage of ListViewWrapper](https://codecov.io/github/pywinauto/pywinauto/pywinauto/controls/common_controls.py?ref=4e8081d053d2a43c8972242dd831ce7341a37794#l-428)).

PR #78, opened 2015-09-17
Workaround for inaccessible submenu.

moden-py:


pywinauto/controls/menuwrapper.py
```diff
@@ -71,6 +71,22 @@ class MenuItemNotEnabled(RuntimeError):
...
 
 
+class MenuInaccessible(RuntimeError):
+    """
+    Raised when a menu has handle but inaccessible
+    """
+    pass
+
+
+def raise_if_inaccessible(method):
```
vasily-v-ryabov: The code structure looks perfect, but I would suggest having `@ensure_accessible` and `self.accessible` names respectively.

moden-py: Thank you, apply  the suggestion.

PR #91, opened 2015-09-27
Fixed button NeedsImageProp property.

moden-py:
Add test for ownerdraw button.
Add tests for static control.

### DIFF
--- a/pywinauto/controls/HwndWrapper.py
+++ b/pywinauto/controls/HwndWrapper.py
@@ -163,6 +163,10 @@ class HwndWrapper(object):
     can_be_label = False
     has_title = True
 
+    # specify whether we need to grab an image of ourselves
+    # when asked for properties
+    _NeedsImageProp = False
+
     #-----------------------------------------------------------
     def __new__(cls, handle):
         # only use the meta class to find the wrapper for HwndWrapper
@@ -207,10 +211,6 @@ class HwndWrapper(object):
         #self.is64bitprocess = handleprops.is64bitprocess(self.ProcessID())
 
         #win32functions.WaitGuiThreadIdle(self)
-
-        # specify whether we need to grab an image of ourselves
-        # when asked for properties
-        self._NeedsImageProp = False
 
         # default to not having a reference control added
         self.ref = None
--- a/pywinauto/unittests/test_win32controls.py
+++ b/pywinauto/unittests/test_win32controls.py
@@ -44,9 +44,13 @@ Timings.window_find_timeout = 3
 Timings.closeclick_dialog_close_wait = .5
 
 mfc_samples_folder = os.path.join(
-   os.path.dirname(__file__), r"..\..\apps\MFC_samples")
+    os.path.dirname(__file__), r"..\..\apps\MFC_samples")
+MFC_tutorial_folder = os.path.join(
+    os.path.dirname(__file__), r"..\..\apps\MFC_tutorial")
+
 if is_x64_Python():
     mfc_samples_folder = os.path.join(mfc_samples_folder, 'x64')
+    MFC_tutorial_folder = os.path.join(MFC_tutorial_folder, 'x64')
 
 
 class ButtonTestCases(unittest.TestCase):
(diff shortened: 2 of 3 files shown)


pywinauto/controls/HwndWrapper.py
```diff
@@ -163,6 +163,10 @@ class HwndWrapper(object):
     can_be_label = False
     has_title = True
 
+    # specify whether we need to grab an image of ourselves
+    # when asked for properties
+    _NeedsImageProp = False
```
airelil: @moden-py, I'm not sure that I understand it. Why do you declare *_NeedsImageProp* as a class attribute, but later it's used as a property of an instance of the class ? You could decorate it as described here: https://www.quora.com/How-do-I-create-a-class-property-in-Python:
```python
@property
def NeedsImageProp():
    return ButtonWrapper._NeedsImageProp
```
However, it seems as not your case. You calculate the value according to instance's *style* variable. Isn't changing the attribute value affects all instances ?

pywinauto/unittests/test_win32controls.py
```diff
@@ -152,8 +159,37 @@ def testIsSelected(self):
         self.assertEquals(self.calc.Radians.GetCheckState(), 1)
 
 
+class ButtonOwnerdrawTestCases(unittest.TestCase):
+
+    """Unit tests for the ComboBoxWrapper(ownerdraw button)"""
```
airelil: Small typo here, should be *ButtonOwnerdrawTestCases*

pywinauto/unittests/test_win32controls.py
```diff
@@ -238,16 +274,6 @@ def testItemData(self):
...
-#
-
-MFC_tutorial_folder = os.path.join(
-   os.path.dirname(__file__), r"..\..\apps\MFC_tutorial")
-if is_x64_Python():
-    MFC_tutorial_folder = os.path.join(MFC_tutorial_folder, 'x64')
-
 
 class ListBoxTestCases(unittest.TestCase):
     "Unit tests for the TreeViewWrapper class"
```
airelil: Ups, one more old typo again, should be *ListBoxTestCases*

pywinauto/unittests/test_win32controls.py
```diff
@@ -647,6 +673,42 @@ def test_menu_handle(self):
...
+
+    def test_NeedsImageProp_ownerdraw(self):
+
+        """test whether an image needs to be saved with the properties"""
+
+        active_window = self.app.active_()
+        self.assertEquals(active_window.Static._NeedsImageProp, True)
+        self.assertIn('Image', active_window.Static.GetProperties())
+
+
```
airelil: Again, if I ended up with *_NeedsImageProp* as a class attribute, I would think about a test with at least two objects of this class to verify that it's been broadcast correctly to all instances or something like that...

pywinauto/controls/HwndWrapper.py
moden-py: @airelil, `_NeedsImageProp = False` in base class used as a default value, expect it would not be changed in subclasses only reassigned directly or via `@property`.I see no reason to use here @property statement. Here is an expected behavior - https://gist.github.com/moden-py/8b62c9e2f7d08bd11474.

pywinauto/unittests/test_win32controls.py
moden-py: Thanks, will fix it

pywinauto/unittests/test_win32controls.py
moden-py: I do not expect an instance or subclass assignment cause baseclass's attribute change.

vasily-v-ryabov: @airelil , do you have a concern about possible usage of `HwndWrapper._NeedsImageProp` access? Right? As I could understand in the example above, the property doesn't broadcast the value to another instances. `_NeedsImageProp` looks like read-only property. So broadcasting looks not possible here.

Or did you say about new test cases required?

PR #98, opened 2015-10-05
docs: try to mock undocumented win32* modules

airelil:
Experimental fix for builds on RTD


airelil:
Added a fix for flaky unittests in HwndWrapper

PR #102, opened 2015-10-13
Added 'cmdline' into _process_get_modules_wmi return

moden-py:


pywinauto/application.py
```diff
@@ -1204,7 +1204,7 @@ def _process_get_modules_wmi():
     # collect all the running processes
     processes = _wmi.ExecQuery('Select * from win32_process')
     for p in processes:
-        modules.append((p.ProcessId, p.ExecutablePath)) # p.Name
+        modules.append((p.ProcessId, p.ExecutablePath, p.CommandLine)) # p.Name
```
vasily-v-ryabov: You need to add `None` in the `process_get_modules` function so that tuple size is the same as in `_process_get_modules_wmi`. Just in case the `process_get_modules` is called somewhere.

moden-py: Thanks, a great idea!

PR #134, opened 2016-01-24
Uia

cetygamer:


dev-requirements.txt
```diff
@@ -2,4 +2,6 @@ pypiwin32
 Pillow==3.0.0
 coverage
 nose
-codecov
\ No newline at end of file
+codecov
+rst2pdf
+Sphinx
```
airelil: Does a user really need rst2pdf & Sphinx ? Aren't they just for developers?

vasily-v-ryabov: The file name is **dev-**requirements.txt. :) This is my experiments with pdf format of the docs. Sphinx html docs is a build artifact now. Pdf docs still fail.

pywinauto/controls/BaseWrapper.py
```diff
@@ -0,0 +1,816 @@
...
+
+    # Properties that describe type of the element
+    can_be_label = False
+    has_title = True
+
+    #------------------------------------------------------------
+    def __new__(cls, elementInfo):
+        # only use the meta class to find the wrapper for BaseWrapper
+        # so allow users to force the wrapper if they want
+        if cls != BaseWrapper:
```
airelil: ```if cls is not BaseWrapper:``` is a bit cleaner (some say it even faster) 
http://stackoverflow.com/questions/132988/is-there-a-difference-between-and-is-in-python
http://jaredgrubb.blogspot.co.nz/2009/04/python-is-none-vs-none.html

dev-requirements.txt
airelil: Ok. I didn't notice it. My mistake.

pywinauto/backend.py
```diff
@@ -0,0 +1,61 @@
...
+    Set active back-end by name
+
+    Possible values of **active_name** are "native", "uia" or
+    other name registered by the **register** function.
+    """
+    if new_active_name not in registered_backends.keys():
+        raise ValueError('Back-end "{backend}" is not registered!'.format(backend=new_active_name))
+
+    #global active_name, ActiveElementInfo, ActiveWrapper
+    global active_name, ActiveElementInfo
```
airelil: QuantifiedCode already complaints about it. Creating a class would be better. Something like this:
```
class Cfg(object):
    active_name = None
    ActiveElementInfo = None

    @classmethod
    def activate(new_active_name):
        ....
cfg = Cfg()
```

pywinauto/__init__.py
```diff
@@ -29,7 +29,12 @@
 
 from . import findwindows
 WindowAmbiguousError = findwindows.WindowAmbiguousError
-WindowNotFoundError = findwindows.WindowNotFoundError
+ElementNotFoundError = findwindows.ElementNotFoundError
+
+from .sysinfo import UIA_support
+if UIA_support:
+    ElementNotFoundError = findwindows.ElementNotFoundError
```
airelil: This statement is a duplicate to the line 32

pywinauto/UIAElementInfo.py
```diff
@@ -20,38 +28,89 @@
...
-        """Create instane of UIAElementInfo from handle.
-        If handle is None create instanse for UI root element"""        
-        if handle is not None:
-            self._element = _iuia.ElementFromHandle(handle)
+    def __init__(self, handle_or_elem = None):
+        """
+        Create instane of UIAElementInfo from a handle (int or long)
+        or from an IUIAutomationElement.
+        If handle_or_elem is None create instanse for UI root element
+        """
```
airelil: Typo errors. Should be "instance", not instane/instanse

PR #138, opened 2016-02-11
Don't enforce warnings.simplefilter('always')

funkyHat:
Prevents turning warnings into exceptions and makes suppressing harder if it's necessary.

Setting the filter within this method is unnecessary and makes overriding it awkward, even if you want to turn warnings into exceptions.

My use-case for suppressing warnings:
I have an application which creates several processes, and I want to find and connect a pywinauto.Application instance to one of those processes. They are easy to find based on window title for which I'm using pywinauto.findwindow.find_windows, but occasionally another application (cmd) will show up as a false match.
This is exacerbated by my having to try in a loop for the window to appear, resulting in many repetitions of the UserWarning being printed.
Of course I can easily check the process name once I've attached the Application, but the warning is already emitted at that point, even though I won't actually try to automate the 64bit cmd.

(I could flip my logic around and attempt to find a matching window belonging to each matching process, but this way looks neater and requires significantly fewer find_windows calls).

### DIFF
--- a/pywinauto/application.py
+++ b/pywinauto/application.py
@@ -88,6 +88,11 @@ class AppNotConnected(Exception):
     pass    #pragma: no cover
 
 
+# Display User and Deprecation warnings every time
+for warning in (UserWarning, PendingDeprecationWarning):
+    warnings.simplefilter('always', warning)
+
+
 #wait_method_deprecation = "Wait* functions are just simple wrappers around " \
 #    "Wait() or WaitNot(), so they may be removed in the future!"
 
@@ -789,7 +794,6 @@ class Application(object):
         the .connect().
         Should be also removed in 0.6.X.
         """
-        warnings.simplefilter('always', PendingDeprecationWarning)
         warnings.warn(
             "connect_()/Connect_() methods are deprecated, "
             "please switch to instance method connect(). "
@@ -848,7 +852,6 @@ class Application(object):
         calling the .start().
         Should be also removed in 0.6.X.
         """
-        warnings.simplefilter('always', PendingDeprecationWarning)
         warnings.warn(
             "start_()/Start_() methods are deprecated, "
             "please switch to instance method start(). "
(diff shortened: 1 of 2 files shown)


vasily-v-ryabov:
Hi Matt, nice to see your efforts improving pywinauto.
The fix is working in py27 and py34+, but [py33 tests](https://ci.appveyor.com/project/pywinauto/pywinauto/build/0.6.0_build340) are failing. Could you please take a look at py33 specific behavior?

funkyHat:
Ok looking at why this failed on 3.3

funkyHat:
So this is weird. Manually testing the same things on 3.3 shows that all the warnings are actually emitted, and also using pytest as a test runner *also* makes them all pass (I found that out because I wanted to use pytest's --pdb feature).

funkyHat:
It looks like nose/unittest in 3.3 fiddles with the warnings filters after loading the modules it's testing, so the filters I add at the top of application.py get overridden.

Playing leapfrog and putting the filters in there again in setUp() seems to work locally. It does mean warnings.filters will keep growing during that test run though so it's not particularly neat...

If you're not happy with that then an alternative option could be:
Running the tests using `py -${PYTHON_VERSION} -W all -m nose` instead of `nosetests` makes them pass too, but makes the test output wordy (all deprecation warnings are now emitted).

Maybe that's not such a bad thing though?

How do you feel about that? I'd rather make the change for all versions but modifying runTestsuite.ps1 to just do that for 3.3 shouldn't be too difficult.

airelil:
Another option is switch to pytest ))

funkyHat:
I almost mentioned that too but thought I'd gone on enough :)
Is there an easy way to get appveyor etc. to run on a branch in my repo or should I just create another PR to test that?

vasily-v-ryabov:
Yeah, I've just thought about pytest too. :)
It's easy to create AppVeyor project for your fork. You just need to copy WebHook URL to the forked repo settings.

vasily-v-ryabov:
Hmm... it looks like `setUp` fix caused [py33 tests](https://ci.appveyor.com/project/pywinauto/pywinauto/build/0.6.0_build343) to pass.
We can merge the PR now. Thank you, Matt!

We may consider porting the whole test suite to pytest later. I think it's interesting idea. But currently we have too many high-priority tasks with 0.6.0 milestone.

funkyHat:
Excellent. Do you want me to create a PR for UIA too, or will you do that?

vasily-v-ryabov:
Master and UIA are not auto mergable now. I'll do that later. Thanks. )

PR #139, opened 2016-02-11
Cross-platform module to emulate mouse events like a real user

MagazinnikIvan:


pywinauto/mouse.py
```diff
@@ -0,0 +1,240 @@
...
+
+
+def move(coords=(0, 0)):
+    "Move the mouse"
+    _perform_click_input(button='move',coords=coords,button_down=False,button_up=False)
+
+
+def press(button='left', coords=(0, 0)):
+    "Press the mouse button"
+    _perform_click_input(button='left', coords=coords, button_down=True, button_up=False)
```
vasily-v-ryabov: Must be `button=button` for every function in the module.

pywinauto/controls/BaseWrapper.py
```diff
@@ -505,16 +540,15 @@ def PressMouseInput(
...
+            button=button,
+            coords=coords,
             button_down=True,
             button_up=False,
             pressed=pressed,
             absolute=absolute,
-            key_down=key_down,
-            key_up=key_up
+#            key_down=key_down,
+#            key_up=key_up
```
vasily-v-ryabov: Please uncomment it. QC warning says `key_up` is unused in the `PressMouseInput` body.

pywinauto/mouse.py
```diff
@@ -0,0 +1,240 @@
...
+            SendKeys.VirtualKeyAction(SendKeys.VK_CONTROL, down=False).Run()
+        if ('shift' in keyboard_keys) and key_up:
+            SendKeys.VirtualKeyAction(SendKeys.VK_SHIFT, down=False).Run()
+        if ('alt' in keyboard_keys) and key_up:
+            SendKeys.VirtualKeyAction(SendKeys.VK_MENU, down=False).Run()
+
+
+else:
+    _display = Display()
+    def _perform_click_input(button='left', coords=(0, 0),
```
vasily-v-ryabov: Linux version of the function has no doc string.

pywinauto/mouse.py
```diff
@@ -0,0 +1,240 @@
...
+    if wheel_dist:
+        _perform_click_input(button='wheel', wheel_dist=wheel_dist, coords=coords)
+
+
+def wheel_click(coords=(0, 0)):
+    "Middle mouse button click at the specified coords"
+    _perform_click_input(button='middle', coords=coords)
+
+
+def _move(coords=(0, 0)):
```
vasily-v-ryabov: No doc string as well.

airelil:
Instead of spoiling the code with endless ```if platform == "win32"```, maybe it's better group platform dependent code in designated sub-modules. Taking http://stackoverflow.com/questions/28551841/how-to-fake-a-module-safely-in-a-python-package?lq=1 as an example we could make something like this:

```
/pywinauto
    __init__.py
        if platform == 'win32'
            #import win32 related modules and functions:
            import win32.win32structures
            import win32.mouse._perform_click
        elif platform == 'Linux'
            #import linux related modules and functions
            import linux.mouse._perform_click
    /win32
        __init__.py
        win32structures
        mouse.py  
    /linux
        __init__.py       
        mouse.py
```

vasily-v-ryabov:
That may be a good idea, but in this particular case we have one platform-dependent internal function and multiple platform-"independent" public wrappers. But for other modules or multi-module packages this structure might be useful.

PR #140, opened 2016-02-22
Back-ends registry re-factoring

vasily-v-ryabov:
- back-end registration is changed (every back-end takes responsibility for that),
 - move `BaseWrapper` to the higher level folder (rename the file to `base_wrapper.py`),
 - separate meta wrappers for Hwnd* and Uia* wrappers,
 - prepare back-ends to be application-specific in the future,
 - fix doc strings and remove some obsolete private methods.

### DIFF
--- a/pywinauto/base_wrapper.py
+++ b/pywinauto/base_wrapper.py
@@ -8,16 +8,16 @@ import win32api
 import win32gui
 import win32process
 import locale
+import abc
 
-from .. import SendKeysCtypes as SendKeys
-from .. import six
-from .. import win32defines, win32structures, win32functions
-from ..timings import Timings
-from ..actionlogger import ActionLogger
-from .. import handleprops
-from .. import mouse
-from ..mouse import _perform_click_input
-from .. import backend
+from . import SendKeysCtypes as SendKeys
+from . import six
+from . import win32defines, win32structures, win32functions
+from .timings import Timings
+from .actionlogger import ActionLogger
+from . import handleprops
+from .mouse import _perform_click_input
+#from . import backend
 
 #=========================================================================
 def remove_non_alphanumeric_symbols(s):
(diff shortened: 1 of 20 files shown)


pywinauto/base_wrapper.py
```diff
@@ -39,79 +40,17 @@ class ElementNotVisible(RuntimeError):
...
-            wrapper = UIAWrapper
-
-        return wrapper
+    def find_wrapper(element):
+        "Abstract static method to find appropriate wrapper"
+        raise NotImplementedError()
 
 #=========================================================================
-@six.add_metaclass(_MetaWrapper)
+@six.add_metaclass(abc.ABCMeta)
```
airelil: I think it has to inherit from ```BaseMeta```

pywinauto/base_wrapper.py
```diff
@@ -39,79 +40,17 @@ class ElementNotVisible(RuntimeError):
...
-        if wrapper_match is None:
-            from .HwndWrapper import HwndWrapper
-            wrapper_match = HwndWrapper
-        return wrapper_match
-
+@six.add_metaclass(abc.ABCMeta)
+class BaseMeta(abc.ABCMeta):
+    "Abstract metaclass for Wrapper objects"
+    
     @staticmethod
```
airelil: Shouldn't you add ```@abc.abstractmethod``` decorator as you proposed in #3   ?
```python
@staticmethod
@abc.abstractmethod
def find_wrapper(element):
    ...
```
Notice it should be applied as the innermost decorator

pywinauto/base_wrapper.py
```diff
@@ -8,19 +8,20 @@
...
+from . import six
+from . import win32defines, win32structures, win32functions
+from .timings import Timings
+from .actionlogger import ActionLogger
+from . import handleprops
+from .mouse import _perform_click_input
+#from . import backend
 
 #=========================================================================
 def remove_non_alphanumeric_symbols(s):
```
airelil: The function handles only space symbols so it should be renamed to ```remove_white_space_symbols``` or changed to support other non-alphanumeric symbols.

airelil:
Maybe we should create a unit test that demonstrates creation of a new backend and tests the behavior of the whole hierarchy.

pywinauto/base_wrapper.py
vasily-v-ryabov: `\W` matches **any** non-alphanumeric character according to `re` module help. Maybe `replace_non_alphanumeric_symbols` is good name as well.

pywinauto/base_wrapper.py
vasily-v-ryabov: It should work in Python3.x but in practice it doesn't work in both Py2 and Py3. I can instantiate `BaseMeta` object despite it's "abstract". See [SO thread](http://stackoverflow.com/questions/4474395/staticmethod-and-abc-abstractmethod-will-it-blend) about that. So I decided to avoid this obsolete construction because I don't believe it's really useful.

pywinauto/base_wrapper.py
vasily-v-ryabov: Good catch! Will fix.

PR #150, opened 2016-03-01
More refactoring.

cetygamer:


pywinauto/controls/HwndWrapper.py
```diff
@@ -191,7 +191,7 @@ def __init__(self, elementInfo):
         if isinstance(elementInfo, six.integer_types):
             elementInfo = NativeElementInfo(elementInfo)
         if hasattr(elementInfo, "_elementInfo"):
-            elementInfo = elementInfo.elementInfo
+            elementInfo = elementInfo._elementInfo
```
vasily-v-ryabov: This introduces QC warning again. I've fixed it by adding read-only property `elementInfo`. Every object (except `self`) must use `element_info` property because it's a public name. Private name must be used inside object only.

Issue #146, opened 2016-02-26
Make backend switching application-specific, not global


vasily-v-ryabov:
Expected result: `app = Application(backend="uia").start('WpfApplication1.exe')`

Internal functions `_get_ctrl`, `_resolve_control` and `_resolve_from_appdata` will become pseudo-private methods of `WindowSpecification`.

Function `findwindows.find_elements` will give 1 more argument: `backend = "native"` (backend name, "native" by default).

PR #151, opened 2016-03-03
Make backends application-specific (#146), fix style warnings

vasily-v-ryabov:


pywinauto/controls/HwndWrapper.py
```diff
@@ -181,19 +178,19 @@ def __new__(cls, element):
...
         is raised.
         """
-        if isinstance(elementInfo, six.integer_types):
-            elementInfo = NativeElementInfo(elementInfo)
-        if hasattr(elementInfo, "_elementInfo"):
-            elementInfo = elementInfo._elementInfo
+        if isinstance(element_info, six.integer_types):
+            element_info = NativeElementInfo(element_info)
+        if hasattr(element_info, "_element_info"):
+            element_info = element_info.element_info
```
airelil: Looks as a missed underscore. Shouldn't it be ```element_info = element_info._element_info``` ?

pywinauto/base_wrapper.py
```diff
@@ -377,7 +372,7 @@ def is_child(self, parent):
     #-----------------------------------------------------------
     def __eq__(self, other):
         "Returns true if 2 BaseWrapper's describe 1 actual element"
-        if hasattr(other, "_elementInfo"):
+        if hasattr(other, "_element_info"):
             return self.element_info == other.element_info
```
airelil: Not sure if I understand it. ``` other``` is tested for  the ```_element_info``` attribute but compared against```other.element_info```

pywinauto/findwindows.py
```diff
@@ -239,10 +242,10 @@ def _title_match(w):
         backup_elements = elements[:]
         elements = []
         for elem in backup_elements:
-            if hasattr(elem, "_elementInfo"):
+            if hasattr(elem, "_element_info"):
                 elements.append(elem.element_info)
```
airelil: Again. test for ```_element_info``` but append ```elem.element_info```. If this is a genuine intention the logic should be explained in comments as it's not obvious from the code

pywinauto/unittests/test_UIAWrapper.py
```diff
@@ -40,10 +40,9 @@ class UIAWrapperTests(unittest.TestCase):
         def setUp(self):
             """Start the application set some data and ensure the application
             is in the state we want it."""
-            backend.activate("uia")
 
             # start the application
-            self.app = Application().Start(os.path.join(wpf_samples_folder, u"WpfApplication1.exe"))
+            self.app = Application(backend = 'uia').Start(os.path.join(wpf_samples_folder, u"WpfApplication1.exe"), timeout = 30)
```
airelil: I see a lot of long lines like that. Why just don't split it:
```python
path_to_exe = os.path.join(wpf_samples_folder, u"WpfApplication1.exe")
self.app = Application(backend = 'uia').Start(path_to_exe, timeout = 30)
```
Even PEP-8 talks about 79 chars per line.

pywinauto/findwindows.py
vasily-v-ryabov: Agree. Will fix everywhere.

pywinauto/unittests/test_UIAWrapper.py
vasily-v-ryabov: I think 79 chars is too old limit because nobody goes to print a code out. Splitting line makes sense to me when it's longer than 100-120 chars. Especially for test code.

vasily-v-ryabov: Ops, this line is really long enough. Will split.

PR #156, opened 2016-03-06
Move DrawOutline into BaseWrapper

moden-py:


pywinauto/unittests/test_HwndWrapper.py
```diff
@@ -319,6 +319,25 @@ def testGetProperties(self):
 #    def testCaptureAsImage(self):
 #        pass
```
moden-py: Disabling new tests since they would not work on AppVeyor environment. "Black screenshots" problem.

airelil:
A small notice, a call ```self.assertEquals``` produces a deprecation warning on py3. Use ```self.assertEqual``` instead.

pywinauto/base_wrapper.py
```diff
@@ -442,6 +443,70 @@ def get_properties(self):
...
+        fill=win32defines.BS_NULL,
+        rect=None):
+        """Draw an outline around the window.
+
+        * **colour** can be either an integer or one of 'red', 'green', 'blue'
+          (default 'green')
+        * **thickness** thickness of rectangle (default 2)
+        * **fill** how to fill in the rectangle (default BS_NULL)
+        * **rect** the coordinates of the rectangle to draw (defaults to
+          the rectangle of the control.
```
vasily-v-ryabov: (minor) right bracket is missing after `control`.

pywinauto/base_wrapper.py
```diff
@@ -442,6 +443,70 @@ def get_properties(self):
     GetProperties = get_properties
 
     #-----------------------------------------------------------
+    def draw_outline(
+        self,
+        colour='green',
```
vasily-v-ryabov: `=` should be wrapped by space according to PEP-8.

pywinauto/base_wrapper.py
```diff
@@ -442,6 +443,70 @@ def get_properties(self):
...
+        thickness=2,
+        fill=win32defines.BS_NULL,
+        rect=None):
+        """Draw an outline around the window.
```
vasily-v-ryabov: should be
```
"""
Draw...
```

pywinauto/base_wrapper.py
moden-py: Vasily, this are the foo arguments. I am sure this not requires spaces around.
...
https://www.python.org/dev/peps/pep-0008/
"Don't use spaces around the = sign when used to indicate a keyword argument or a default parameter value. "

vasily-v-ryabov: Okay, I mixed it up with another case.

PR #157, opened 2016-03-12
Use IUIAutomationPropertyCondition filters for children/descendants search in UIA backend

airelil:
Turn ElementInfo.children and ElementInfo.descendants into methods to allow passing the optional filtering arguments down to the backend specific search function.


pywinauto/UIAElementInfo.py
```diff
@@ -167,29 +167,46 @@ def parent(self):
...
+        "Find all elements according to the given scope and conditions"
+        elements = []
+        ptrs_array = self._element.FindAll(scope, cond)
+        for num in range(ptrs_array.Length):
+            child = ptrs_array.GetElement(num)
+            elements.append(UIAElementInfo(child))
+
+        return elements
+
+    def _build_condition(self, process = None, class_name = None, title = None):
```
vasily-v-ryabov: Often QC finds false alarms as "new issues" but now I agree with QC. `self` is not used here so it might be function or static method. As far as I remember, pseudo-private methods should start with `__` (double underscore). Single one is used for functions.

pywinauto/UIAElementInfo.py
```diff
@@ -167,29 +167,46 @@ def parent(self):
...
+        if title:
+            # TODO: CreatePropertyConditionEx with PropertyConditionFlags_IgnoreCase
+            new_cond = _iuia.CreatePropertyCondition(
+                                    _UIA_dll.UIA_NamePropertyId, title)
+            full_cond = _iuia.CreateAndCondition(new_cond, full_cond)
+
+        return full_cond
+
+    def children(self, **kwargs):
+        "Return a list of only immediate children of the element according to the criterias"
```
vasily-v-ryabov: typo: "criteria" is already plural. BTW, triple double quotes are preferable by landscape.io for a doc string.

pywinauto/base_wrapper.py
```diff
@@ -517,7 +518,7 @@ def is_child(self, parent):
         element if the parent element is the the chain of parent elements
         for the child element.
         """
-        return self in parent.children()
+        return self in parent.children(class_name = self.class_name())
```
vasily-v-ryabov: Good! This should really boost the method. Not so obviously first, but I got it.

pywinauto/unittests/test_UIAWrapper.py
```diff
@@ -60,6 +60,36 @@ def testFriendlyClassName(self):
             "Test getting the friendly classname of the dialog"
             self.assertEqual(self.button.friendly_class_name(), "Button")
 
+        def testFindNonTopCtlByClassNameAndTitle(self):
```
vasily-v-ryabov: It's not so critical to have all test methods PEP-8 complaint. But I think new methods might be in PEP-8. As well as doc strings.

pywinauto/UIAElementInfo.py
airelil: Not exactly. The single underscore is for non-public interfaces of any kind: attributes, stand-alone functions. It just means the interface is for the internal usage only 

> 

> Non-public attributes are those that are not intended to be used by third parties; you make no guarantees that non-public attributes won't change or even be removed.

 The double forward underscore, as you said, is a sort of "sealing" of the method and preventing its sub-classing. Not sure that's the goal here. Basically, UIAElementInfo isn't supposed to be sub-classed but we do not prohibit it.

PR #159, opened 2016-03-16
Make UIA staff singleton, optimize Kill_() method, style fixes

vasily-v-ryabov:
[x] Make UIA staff singleton in `uia_defines.py`
[x] Use CreateAndConditionFromArray
[x] optimize Kill_() method, add `force_close()` method to `DialogWrapper`
[x] fix several doc strings and style warnings


pywinauto/uia_defines.py
```diff
@@ -42,7 +106,8 @@
 # header: UIAutomationClient.h
 
 def _build_pattern_ids_dic():
-    """A helper procedure to build a registry of control patterns 
+    """
```
airelil: We need to agree about the style. PEP-8 sends to PEP-257 and the latter says that it can be trimmed to a single line. So I started fixing comments to be on the same line with the quotes and as I see you prefer to keep it separated.

airelil:
A lot of style fixes, so hard to see the main change. Not sure if the singleton is needed, but I liked refactoring with moving common things in uia_define.

pywinauto/uia_defines.py
vasily-v-ryabov: I'm just trying to reduce number of landscape.io warnings. I'd suggest to follow both QC and Landscape recommendations. Landscape says that multi line comment should have 1 line below quotes, 1 blank line and then more detailed description. Closing quotes should stay on separate line as well. For one line doc string every thing is at one line.

vasily-v-ryabov:
Yeah, full diff looks too long. Maybe it's better to review every commit, but some commits have incorrect things that are already fixed later.

PR #161, opened 2016-03-19
UIA ComboBox wrapper

airelil:
- Add the WpfApplication1.exe with ComboBox control
- Add UIA ComboBox wrapper and unit tests
- PEP-8 style fixes


pywinauto/controls/UIAWrapper.py
```diff
@@ -250,6 +249,127 @@ def invoke(self):
...
+        a control might not have any items selected by default, 
+        meaning that IsSelectionRequired is FALSE. However, 
+        after an item is selected the control must always have 
+        at least one item selected.
+        """
+        elem = self.element_info.element
+        iface = uia_defs.get_elem_interface(elem, "Selection")
+        return iface.CurrentIsSelectionRequired
+
+    def select_by_name_or_by_idx(self, item_name = None, item_index = 0):
```
vasily-v-ryabov: Maybe more simple name `select` is better here. It's also aligned with the pattern method name.

pywinauto/controls/UIAWrapper.py
```diff
@@ -250,6 +249,127 @@ def invoke(self):
...
+        ComboBox, TreeView, ListView
+        """
+
+        list_ = self.element_info.children(title = item_name)
+        if item_index < len(list_):
+            elem = list_[item_index].element
+            iface = uia_defs.get_elem_interface(elem, "SelectionItem")
+            iface.Select()
+        else:
+            raise ValueError("item not found")
```
vasily-v-ryabov: `ValueError` is fine for not found `item_name`, but `IndexError` is better for `item_index` case. Allowing both params for the method means that we need to search by title_re (maybe). I'd suggest to have 1 param `item` only as it was done in `ListViewWrapper`. The `item` param can be interpreted as a string or as an item. We just need to check for its type: `int` or not.

pywinauto/controls/uia_controls.py
```diff
@@ -131,3 +135,101 @@ def is_selected(self):
         elem = self.element_info.element
         iface = uia_defs.get_elem_interface(elem, "SelectionItem")
         return iface.CurrentIsSelected
+
+#====================================================================
+class ComboBoxWrapper(UIAWrapper.UIAWrapper):
+    """Wrap a UIA CoboBox control"""
```
vasily-v-ryabov: (minor) Landscape.io recommends to have an empty line before a class doc string and after it.

pywinauto/controls/uia_controls.py
```diff
@@ -131,3 +135,101 @@ def is_selected(self):
...
+
+    #-----------------------------------------------------------
+    def __init__(self, hwnd):
+        """Initialize the control"""
+        super(ComboBoxWrapper, self).__init__(hwnd)
+
+    #-----------------------------------------------------------
+    def texts(self):
+        """Return the text of the items in the combobox"""
+
```
vasily-v-ryabov: (minor) Sorry again, but Landscape recommends to remove an empty line after a function/method doc string.

pywinauto/uia_defines.py
```diff
@@ -158,15 +158,20 @@ def _build_pattern_ids_dic():
 # The definition can also be found in the comtypes package 
 # In a file automatically generated according to UIAutomation GUID:
 # comtypes\gen\_944DE083_8FB8_45CF_BCB7_C477ACB2F897_*.py
-toggle_state_off = 0
-toggle_state_on = 1
-toggle_state_inderteminate = 2
+toggle_state_off = IUIA().ui_automation_client.ToggleState_Off
+toggle_state_on = IUIA().ui_automation_client.ToggleState_On
```
vasily-v-ryabov: This is much better than magic numbers. :+1:

pywinauto/controls/uia_controls.py
```diff
@@ -31,7 +31,6 @@
 """
 Wrap various UIA windows controls
 """
-import pywinauto.six as six
 import pywinauto.uia_defines as uia_defs
```
vasily-v-ryabov: I'd prefer having only relative imports everywhere except tests (like `from .. import uia_defines as uia_defs`).

PR #163, opened 2016-03-22
Added UIA EditWrapper class

cetygamer:


pywinauto/controls/uia_controls.py
```diff
@@ -219,3 +228,199 @@ def item_count(self):
...
+            search_range = document_range.FindText(string_to_select, False, False)
+
+            try:
+                search_range.Select()
+            except ValueError:
+                raise RuntimeError("Text '{0}' hasn't been found".format(string_to_select))
+
+            # give the control a chance to catch up before continuing
+            if self.element_info.handle:
+                win32functions.WaitGuiThreadIdle(self)
```
vasily-v-ryabov: I'm not sure that waiting for something is really necessary here. Probably `Select()` method of the pattern should be a blocking call. So returning from it means all actions are done, isn't it?

pywinauto/controls/uia_controls.py
```diff
@@ -32,6 +32,15 @@
...
 from .. import uia_defines as uia_defs
+
+import time
+import locale
+
+from .. import six
+from ..timings import Timings
+from .. import win32functions
+
+from .. import uia_defines as uia_defs
```
vasily-v-ryabov: `uia_defs` is already imported. Please remove duplicated import.

PR #164, opened 2016-03-23
Make CloseClick() more stable and fix style warnings

vasily-v-ryabov:


pywinauto/application.py
```diff
@@ -1212,7 +1212,7 @@ def process_get_modules():
     # collect all the running processes
     pids = win32process.EnumProcesses()
     for pid in pids:
-        if pid != 0: # skip system process (0x00000000)
+        if pid != 0 and isinstance(pid, int): # skip system process (0x00000000)
```
airelil: Just curious, how come that  ```win32process.EnumProcess``` returns non-integer values. And even if that happens ```process_module``` should raise a validation error on the bad ```pid``` parameter.
BTW, maybe it's better to filter out the system process before the loop:
```python
pids = list(win32process.EnumProcesses())
pids.remove(0)
for pid in pids:
    try:
         modules.append((pid, process_module(pid), None))
    except (win32gui.error, ProcessNotFoundError):
        continue
```

vasily-v-ryabov: This is just for peace of mind. I've done this for the following test instability:
```python
ERROR: Test minimizing a sample app into the hidden area of the tray

  File "C:\projects\pywinauto-8b7t7\pywinauto\unittests\test_taskbar.py", line 156, in tearDown
    app.connect(path="TrayMenu.exe")
  File "C:\projects\pywinauto-8b7t7\pywinauto\application.py", line 864, in connect
    self.__warn_incorrect_bitness()
  File "C:\projects\pywinauto-8b7t7\pywinauto\application.py", line 959, in __warn_incorrect_bitness
    if self.is64bit() != is_x64_Python():
  File "C:\projects\pywinauto-8b7t7\pywinauto\application.py", line 974, in is64bit
    return handleprops.is64bitprocess(self.process)
  File "C:\projects\pywinauto-8b7t7\pywinauto\handleprops.py", line 152, in is64bitprocess
    phndl = win32api.OpenProcess(win32con.MAXIMUM_ALLOWED, 0, process_id)
pywintypes.error: (87, 'OpenProcess', 'The parameter is incorrect.')
```
Not absolutely sure, that a non-integer pid is the root cause.

PR #177, opened 2016-04-14
LVITEMW support for 64-bit python + 32-bit app

KalyakulinaAlena:


airelil:
Too many changes in unit tests. I'd go with inheritance of test suite and running it with a different parameter when it's needed.
```python
from __future__ import print_function
import unittest

# check if we need x64 tests
x64 = True

class Win32Tests(unittest.TestCase):
   """ Tests for x86"""
    path = "win32.exe"
    def setUp(self):
        print("setUp", self.path)

    def tearDown(self):
        print("tearDown")

    def testCall(self):
        print("testCall", (self.path))

if x64:
    class Win64Tests(Win32Tests):
        """ Same tests but for x64"""
        path = "win64.exe"

if __name__ == "__main__":
    unittest.main()
```

Issue #85, opened 2015-09-21
Need to test an appdata capabilities for localization independent scripting


vasily-v-ryabov:
There is a [sample MFC application with multiple language support](http://www.codeproject.com/Articles/11040/Multiple-language-support-for-MFC-applications-wit). We need a test that using `match_history` and `appdata` attributes to save and load English texts of the controls and re-run the same part of the script for another language. Creating an example in `.\examples\` folder would also be very useful.

PR #179, opened 2016-04-16
PEP-8 / PEP-257 fixes and better coverage for application.py

vasily-v-ryabov:
* [x] `cpu_usage()` and `process_get_modules()` have been additionally covered.
* [x] Renamed`WrapperObject` -> `wrapper_object` etc.

### DIFF
--- a/pywinauto/application.py
+++ b/pywinauto/application.py
@@ -1283,23 +1283,16 @@ def process_from_module(module):
         modules = _process_get_modules_wmi()
     except Exception:
         modules = process_get_modules()
-    
-    accessible_modules = process_get_modules()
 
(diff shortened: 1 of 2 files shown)


pywinauto/application.py
```diff
@@ -1135,24 +1156,24 @@ def __getattribute__(self, attr_name):
         return self[attr_name]
 
     def WriteAppData(self, filename):
```
airelil: Maybe it's time to convert it to non-public method: ```_write_app_data```. Same for the GetMatchHistory method.

vasily-v-ryabov: I think it's a part of issue #85. I want to review this functionality and make some re-factoring simultaneously.

PR #181, opened 2016-04-16
Refactor tests for LVITEM

KalyakulinaAlena:
@vasily-v-ryabov Rewrite tests with inheritance. Please, review

### DIFF
--- a/pywinauto/controls/common_controls.py
+++ b/pywinauto/controls/common_controls.py
@@ -675,7 +675,7 @@ class ListViewWrapper(HwndWrapper.HwndWrapper):
         if self.is_unicode():
             self.create_buffer = ctypes.create_unicode_buffer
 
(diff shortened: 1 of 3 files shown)


pywinauto/controls/common_controls.py
```diff
@@ -674,15 +674,27 @@ def __init__(self, hwnd):
 
         if self.is_unicode():
             self.create_buffer = ctypes.create_unicode_buffer
-            self.LVCOLUMN       = win32structures.LVCOLUMNW
-            self.LVITEM         = win32structures.LVITEMW
+
+            if is64bitprocess(hwnd.process_id) or not is_x64_Python():
```
vasily-v-ryabov: `hwnd.process_id` should be replaced with `self.process_id()` as it was done by Igor.

PR #187, opened 2016-04-25
send_chars method implementation

Syainova-Diana:
@vasily-v-ryabov 
Could you help me with my Task №2, please? 

I have implemented a simple version of `send_chars` and have questions about processing the sequence "{TAB}{ENTER 2}" and modifiers Ctrl(^), Alt(%) and Shift (+).

1. How should I handle this sequence and modifiers?
2. I could not find **Alt** in `pywinauto.SendKeysCtypes` . Should I replace the **Alt** treatment with the treatment **Menu**?

### DIFF
--- a/pywinauto/controls/HwndWrapper.py
+++ b/pywinauto/controls/HwndWrapper.py
@@ -476,7 +476,6 @@ class HwndWrapper(BaseWrapper):
         """
 
         keys = SendKeysCtypes.parse_keys(message, with_spaces, with_tabs, with_newlines)
-        res=[]
         for key in keys:
 
             vk, scan, flags = key.get_key_info()
@@ -486,14 +485,9 @@ class HwndWrapper(BaseWrapper):
             if isinstance(key, SendKeysCtypes.VirtualKeyAction):
 
                 if key.down and key.up and (flags == 0):
+                    # TODO: {RETURN} virtual key doesn't work
                     lparam = 1 << 0 | scan << 16
-                    if vk == win32con.VK_RETURN:
-                        # + Enter
-                        win32gui.PostMessage(self.handle, win32con.WM_KEYDOWN, vk, lparam)
-                        win32gui.PostMessage(self.handle, win32con.WM_KEYUP, vk, lparam)
-                    else:
-                        # + BACKSPACE
-                        win32api.SendMessage(self.handle, win32con.WM_CHAR, vk, lparam)
+                    win32api.SendMessage(self.handle, win32con.WM_CHAR, vk, lparam)
                 elif key.down and key.up and (flags == 1):
                     # + LEFT, DELETE
                     lparam = 1 << 0 | scan << 16 | flags << 24 | 0 << 29 | 0 << 31
@@ -501,13 +495,16 @@ class HwndWrapper(BaseWrapper):
                     lparam = 1 << 0 | scan << 16 | flags << 24 | 0 << 29 | 1 << 30 | 1 << 31
                     win32api.SendMessage(self.handle, win32con.WM_KEYUP, vk, lparam)
                 elif key.down:
+                    # TODO: {CTRL} (^) modifier doesn't work
                     # + SHIFT down
                     # - CTRL down
+                    # print('key', key, 'down')
                     lparam = 1 << 0 | scan << 16 | flags << 24 | 0 << 29 | 0 << 31
                     win32api.SendMessage(self.handle, win32con.WM_KEYDOWN, vk, lparam)
                 elif key.up:
                     # + SHIFT up
                     # - CTRL up
+                    # print('key', key, 'up')
                     lparam = 1 << 0 | scan << 16 | flags << 24 | 0 << 29 | 1 << 30 | 1 << 31
                     win32api.SendMessage(self.handle, win32con.WM_KEYUP, vk, lparam)
             elif isinstance(key, SendKeysCtypes.EscapedKeyAction):
@@ -518,10 +515,7 @@ class HwndWrapper(BaseWrapper):
                 # Usual key
                 win32api.SendMessage(self.handle, win32con.WM_CHAR, scan, lparam)
 
-            time.sleep(0.05)
-
-            res.append([key, vk, scan, flags])
-        return res
+            # time.sleep(Timings.after_sendkeys_key_wait)
 
     #-----------------------------------------------------------
     def send_message_timeout(
--- a/pywinauto/controls/HwndWrapper.py
+++ b/pywinauto/controls/HwndWrapper.py
@@ -478,22 +478,14 @@ class HwndWrapper(BaseWrapper):
         keys = SendKeysCtypes.parse_keys(message, with_spaces, with_tabs, with_newlines)
 
         for key in keys:
-            #keyinfo = key._get_key_info()
-            #res.append([key, keyinfo])
-
-            if (isinstance(key, SendKeysCtypes.EscapedKeyAction) or
-                    isinstance(key, SendKeysCtypes.VirtualKeyAction)):
-                key.Run()
-            else:
-                win32api.SendMessage(self.handle, win32con.WM_CHAR, key.get_key_scan_code(), 0)
-
+            key.SendToHandle(self.handle)
             time.sleep(0.05)
 
     #-----------------------------------------------------------
     def send_message_timeout(
         self,
         message,
-        wparam = 0 ,
+        wparam = 0,
         lparam = 0,
         timeout = None,
         timeoutflags = win32defines.SMTO_NORMAL):
--- a/pywinauto/controls/HwndWrapper.py
+++ b/pywinauto/controls/HwndWrapper.py
@@ -463,31 +463,31 @@ class HwndWrapper(BaseWrapper):
     SendMessage = send_message
 
     # -----------------------------------------------------------
-    def send_chars(self, message):
-        """Send a string to the control and wait for it to return"""
-        res = []
+    def send_chars(self,
+                   message,
+                   with_spaces=True,
+                   with_tabs=True,
+                   with_newlines=True):
+        """
+        Silently send a string to the control
 
-        mspl = message.split("{TAB}{ENTER 2}")
-        ind = 0
-        for substr in mspl:
-            ind = ind + 1
+        Parses modifiers Shift(+), Control(^), Menu(%) and Sequences like "{TAB}", "{Enter}"
+        For more information about Sequences and Modifiers navigate to SendKeysCtypes.py
+        """
 
-            for c in substr:
-                if c in SendKeysCtypes.MODIFIERS.keys():
-                    # Shift, Ctrl, Menu
-                    res.append(-1)
-                else:
-                    # Usual character
-                    res.append(win32api.SendMessage(self.handle, win32con.WM_CHAR, ord(c), 0))
+        keys = SendKeysCtypes.parse_keys(message, with_spaces, with_tabs, with_newlines)
 
-            if ind != len(mspl):
-                # "{TAB}{ENTER 2}"
-                res.append(-2)
+        for key in keys:
+            #keyinfo = key._get_key_info()
+            #res.append([key, keyinfo])
 
-        return res
+            if (isinstance(key, SendKeysCtypes.EscapedKeyAction) or
+                    isinstance(key, SendKeysCtypes.VirtualKeyAction)):
+                key.Run()
+            else:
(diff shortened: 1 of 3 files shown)


vasily-v-ryabov:
@Syainova-Diana probably `parse_keys` function is what you need. You may add public method to get key code (since `_get_key_info` is a private one) for every `*Key*` class because you send `WM_CHAR` instead of calling `Run()`.

vasily-v-ryabov:
2) Yep, `'%': VK_MENU,` seems the same as `Alt`.

pywinauto/controls/HwndWrapper.py
```diff
@@ -461,6 +462,33 @@ def send_message(self, message, wparam = 0, lparam = 0):
...
+                    res.append(win32api.SendMessage(self.handle, win32con.WM_CHAR, ord(c), 0))
+
+            if ind != len(mspl):
+                # "{TAB}{ENTER 2}"
+                res.append(-2)
+
+        return res
+
+    # Non PEP-8 alias
+    SendChars = send_chars
```
vasily-v-ryabov: Old-style alias is not necessary for new methods. Please leave `send_chars` only.

pywinauto/controls/HwndWrapper.py
```diff
@@ -461,6 +462,33 @@ def send_message(self, message, wparam = 0, lparam = 0):
     # Non PEP-8 alias
     SendMessage = send_message
 
+    # -----------------------------------------------------------
+    def send_chars(self, message):
+        """Send a string to the control and wait for it to return"""
```
vasily-v-ryabov: Should be like that:
```python
"""
Silently send a string to the control

More detailed description with possible modifiers and its meaning.
"""
```

pywinauto/unittests/test_HwndWrapper.py
```diff
@@ -276,6 +276,14 @@ def testSendMessage(self):
         expected = 0x89 # 0x2000 + 0x40
         self.assertEqual(expected, code)
 
+    def testSendChars(self):
```
vasily-v-ryabov: should be `test_send_chars`

Syainova-Diana:
I can't use `{Enter}` in Test because it closes the application.

pywinauto/controls/HwndWrapper.py
```diff
@@ -463,31 +463,31 @@ def send_message(self, message, wparam = 0, lparam = 0):
...
-        return res
+            if (isinstance(key, SendKeysCtypes.EscapedKeyAction) or
+                    isinstance(key, SendKeysCtypes.VirtualKeyAction)):
+                key.Run()
+            else:
+                win32api.SendMessage(self.handle, win32con.WM_CHAR, key.get_key_scan_code(), 0)
 
-    # Non PEP-8 alias
-    SendChars = send_chars
+            time.sleep(0.05)
```
vasily-v-ryabov: Magic number should be replaced by `Timings.<some_timing_value>` variable that can be controlled by `timings.py`. See `type_keys` implementation for reference. You can create new timing variable or re-use one from `type_keys`.

pywinauto/controls/HwndWrapper.py
```diff
@@ -463,31 +463,31 @@ def send_message(self, message, wparam = 0, lparam = 0):
...
+                   with_tabs=True,
+                   with_newlines=True):
+        """
+        Silently send a string to the control
 
-        mspl = message.split("{TAB}{ENTER 2}")
-        ind = 0
-        for substr in mspl:
-            ind = ind + 1
+        Parses modifiers Shift(+), Control(^), Menu(%) and Sequences like "{TAB}", "{Enter}"
```
vasily-v-ryabov: I'd prefer describing it as `Alt(%)`. Maybe adding alias `VK_ALT = VK_MENU` to `SendKeysCtypes.py` could be less confusing.

vasily-v-ryabov:
`{ENTER}` is just an example. We need to test at least 2 sequences like `{TAB 2}` or `{BACKSPACE}` or something else.

pywinauto/controls/HwndWrapper.py
```diff
@@ -461,6 +462,33 @@ def send_message(self, message, wparam = 0, lparam = 0):
...
+
+        for key in keys:
+            #keyinfo = key._get_key_info()
+            #res.append([key, keyinfo])
+
+            if (isinstance(key, SendKeysCtypes.EscapedKeyAction) or
+                    isinstance(key, SendKeysCtypes.VirtualKeyAction)):
+                key.Run()
+            else:
+                win32api.SendMessage(self.handle, win32con.WM_CHAR, key.get_key_scan_code(), 0)
```
vasily-v-ryabov: `lparam` is not used here, but it's important to send an info about modifiers etc. See [MSDN docs](https://msdn.microsoft.com/en-us/library/windows/desktop/ms646276(v=vs.85).aspx).

pywinauto/controls/HwndWrapper.py
```diff
@@ -461,6 +462,33 @@ def send_message(self, message, wparam = 0, lparam = 0):
...
+
+        keys = SendKeysCtypes.parse_keys(message, with_spaces, with_tabs, with_newlines)
+
+        for key in keys:
+            #keyinfo = key._get_key_info()
+            #res.append([key, keyinfo])
+
+            if (isinstance(key, SendKeysCtypes.EscapedKeyAction) or
+                    isinstance(key, SendKeysCtypes.VirtualKeyAction)):
+                key.Run()
```
vasily-v-ryabov: `key.Run()` is strictly prohibited here, because it's not silent like the whole `type_keys`. We need to make difference, not a copy of `type_keys`.

pywinauto/unittests/test_HwndWrapper.py
```diff
@@ -276,6 +276,14 @@ def testSendMessage(self):
         expected = 0x89 # 0x2000 + 0x40
         self.assertEqual(expected, code)
 
+    def test_send_chars(self):
+        testString = "+hello +world"
```
vasily-v-ryabov: We also need to call `dlg.minimize()` to test it in more realistic conditions, because `send_chars` is supposed to be silent (i.e. working for even minimized window).

Syainova-Diana:
@vasily-v-ryabov 
Where can I find **lparam** and how to handle **EscapedKeyAction**?

vasily-v-ryabov:
@Syainova-Diana, please revert back the recent commit. We don't need to add `SendToHandle`. This is isolated functionality for `HwndWrapper`.

The `SendMessage/send_message` signature is the following:
```python
send_message(hwnd, msg_id, wParam, lParam)
```
So every message description (in MSDN) contains `wParam` and `lParam` possible values. You pass `0` as `lParam` for `WM_CHAR`. But [in the MSDN docs](https://msdn.microsoft.com/en-us/library/windows/desktop/ms646276(v=vs.85).aspx) there are some possible non-zero values that are helpful to process modifiers like `Alt`, `Ctrl` and `Shift`.

For `WM_CHAR` message `lParam` is an integer with possible bit values. This is an example how to set 24-th and 29-th bits in integer `x`:
```python
x = 1 << 24 | 1 << 29
# check that it works
>>> bin(x)
'0b100001000000000000000000000000'
```

Syainova-Diana:
@vasily-v-ryabov 
But i'm not sure if it works with all sequences as well as with **{DEL}** and **{LEFT}**.

pywinauto/controls/HwndWrapper.py
```diff
@@ -461,11 +462,50 @@ def send_message(self, message, wparam = 0, lparam = 0):
...
+
+            vk, scan, flags = key.get_key_info()
+
+            lparam = scan << 16 | flags << 24
+
+            if isinstance(key, SendKeysCtypes.VirtualKeyAction):
+                # Shift, Left, Back, Delete, ...
+                if key.down:
+                    lparam = lparam | flags << 24 | 0 << 29 | 0 << 31
+                    win32api.SendMessage(self.handle, win32con.WM_KEYDOWN, vk, lparam)
```
vasily-v-ryabov: `WM_KEYDOWN` and `WM_KEYUP` should NOT work for minimized window (because it's the same that `type_keys` does). Maybe you mean `WM_CHAR` too in this case?

vasily-v-ryabov:
I see it works for `{DEL}` and `{LEFT}` for now but `{ENTER}` and `{BACKSPACE}` don't work. Probably `WM_KEYDOWN` and `WM_KEYUP` must be replaced with `WM_CHAR`.

vasily-v-ryabov:
OK, after short debugging I could get it to work if the following code is running for `VirtualKeyAction`:
```python
            if isinstance(key, SendKeysCtypes.VirtualKeyAction):
                # Shift, Left, Back, Delete, ...
                lparam = lparam | 1 << 0 | flags << 24 | 0 << 29 | 1 << 30 | 0 << 31
                win32api.SendMessage(self.handle, win32con.WM_CHAR, vk, lparam)
```

vasily-v-ryabov:
But the following sequence is still not working: `ctrl.send_chars('^a^c{RIGHT}^v')` Maybe some manipulation with 30th and 31st bits is necessary.

Syainova-Diana:
3 times beacuse of **{DEL 3}**

Syainova-Diana:
@vasily-v-ryabov 
Manipulation with 30th and 31st bits is not helping

vasily-v-ryabov:
I'd suggest to stop at the following code:
```python
        for key in keys:

            vk, scan, flags = key.get_key_info()

            lparam = 1 << 0 | scan << 16 | flags << 24
            #print('key =', key, ', flags =', flags)

            if isinstance(key, SendKeysCtypes.VirtualKeyAction):

                if key.down and key.up: # and (flags == 0):
                    lparam = 1 << 0 | scan << 16
                    win32api.SendMessage(self.handle, win32con.WM_CHAR, vk, lparam)
                elif key.down and key.up and (flags == 1):
                    # TODO: {LEFT} and {RIGHT} virtual keys don't work
                    # + LEFT, DELETE
                    #print('[LEFT, DELETE] key', key, 'press, flags =', flags)
                    lparam = 1 << 0 | scan << 16 | flags << 24 | 0 << 29 | 0 << 30 | 0 << 31
                    win32api.SendMessage(self.handle, win32con.WM_CHAR, vk, lparam)
                    lparam = 1 << 0 | scan << 16 | flags << 24 | 0 << 29 | 1 << 30 | 1 << 31
                    win32api.SendMessage(self.handle, win32con.WM_CHAR, vk, lparam)
                elif key.down:
                    # TODO: {CTRL} (^) modifier doesn't work
                    # + SHIFT down
                    # - CTRL down
                    #print('key', key, 'down')
                    lparam = 1 << 0 | scan << 16 | flags << 24 | 0 << 29 | 0 << 30 | 0 << 31
                    win32api.SendMessage(self.handle, win32con.WM_CHAR, vk, lparam)
                elif key.up:
                    # + SHIFT up
                    # - CTRL up
                    #print('key', key, 'up')
                    lparam = 1 << 0 | scan << 16 | flags << 24 | 0 << 29 | 1 << 30 | 1 << 31
                    win32api.SendMessage(self.handle, win32con.WM_CHAR, vk, lparam)
            elif isinstance(key, SendKeysCtypes.EscapedKeyAction):
                # An escaped key action e.g. F9 DOWN, etc
                # And key between Shifts. a -> A
                win32api.SendMessage(self.handle, win32con.WM_CHAR, vk, lparam)
            else:
                # Usual key
                win32api.SendMessage(self.handle, win32con.WM_CHAR, scan, lparam)

            #time.sleep(Timings.after_sendkeys_key_wait)
```

Syainova-Diana:
@vasily-v-ryabov 
Left and Right didn't work because of

> #print('[LEFT, DELETE] key', key, 'press, flags =', flags)
>                     lparam = 1 << 0 | scan << 16 | flags << 24 | 0 << 29 | 0 << 30 | 0 << 31
>                     win32api.SendMessage(self.handle, win32con.**WM_CHAR**, vk, lparam)
>                     lparam = 1 << 0 | scan << 16 | flags << 24 | 0 << 29 | 1 << 30 | 1 << 31
>                     win32api.SendMessage(self.handle, win32con.**WM_CHAR**, vk, lparam)

It works with **WM_KEYDOWN** and **WM_KEYUP**

vasily-v-ryabov:
@Syainova-Diana, OK, let's use `WM_KEY*` messages for that. It seems working for minimized window.

vasily-v-ryabov:
By the way, `{ENTER}` is working for me. Just `Ctrl+<key>` combinations still fail.

Issue #43, opened 2015-06-18
Implement keyboard and mouse hooks module (say hooks.py)


vasily-v-ryabov:
pyHook contains C module. But hooks can be implemented with Python only (using ctypes).

See pyHook example for reference: http://stackoverflow.com/questions/4581772/using-pyhook-to-respond-to-key-combination-not-just-single-keystrokes

One more example of keyboard hook without pyHook: http://stackoverflow.com/questions/9817531/applying-low-level-keyboard-hooks-with-python-and-setwindowshookexa

Analogue in AutoIt:

```
HotKeySet('{F2}', '_hk1')

While 1
Sleep(100)
WEnd

Func _hk1()
Send("{CTRLDOWN}{SHIFTDOWN}t{SHIFTUP}{CTRLUP}")
EndFunc
```

Thanks [TomashUA](https://toster.ru/user/TomashUA) for the use case.

vasily-v-ryabov:
One more Python module to look at: https://github.com/IronManMark20/hooked

vasily-v-ryabov:
Another alternative: https://github.com/schurpf/pyhk

They are all for keyboard only. We need mouse hooks as well.

airelil:
@vasily-v-ryabov, just to clarify. Do you mean mouse/keyboard hooks for processes started/connected to by *pywinauto* ? The AutoIt example above sets the hook for it's own process. It's a bit different.
In case we are hooking another process we have to check if there are special requirements for that: administrator rights, code signing...

vasily-v-ryabov:
Low-level mouse and keyboard hooks are global. We may set it in operating system and then filter target process events in. I didn't check yet how it works for processes with different privileges.

Code signing is another interesting topic that we didn't touch at all.

airelil:
Maybe you can put some use cases to better understand the requirements? I could think about hooking of Python interpreter process itself with a running Pywinauto script. (Similar to the specified AutoId example) This could allow to interrupt/pause/whatever the script execution in more easier manner, without trying to the catch interpreter window. However, I still don't see how it can be of use for other processes.

vasily-v-ryabov:
Currently we have only one example using the outdated pyHook: [recorder.py](https://github.com/pywinauto/pywinauto/blob/master/sandbox/recorder.py). Since mouse/keyboard hooks are global (i.e. system wide), we can get a window handle from any process (in theory) in the event callback. It means that our Python interpreter instance listens all processes with allowed permission (at least).

[...]

PR #188, opened 2016-04-26
hooks.py module (implement #43)

MaxSamokhvalov:
hooks.py was added


pywinauto/hooks.py
```diff
@@ -0,0 +1,231 @@
+from ctypes import wintypes
+from ctypes import windll, CFUNCTYPE, POINTER, c_int, c_void_p, byref
+import atexit
+
+
+def create_pointer(handler):
```
vasily-v-ryabov: `_create_callback_pointer` or just `_callback_pointer` would be more informative. It seems we don't need to share this function with other users so let's keep it private.

pywinauto/hooks.py
```diff
@@ -0,0 +1,231 @@
...
+    cmp_func = CFUNCTYPE(c_int, c_int, c_int, POINTER(c_void_p))
+    return cmp_func(handler)
+
+
+class KeyboardEvent:
```
vasily-v-ryabov: New-style classes are recommended to use everywhere, i.e. just inherit all classes from `object`.

pywinauto/hooks.py
```diff
@@ -0,0 +1,231 @@
...
+                return windll.user32.CallNextHookEx(self.id, code, event_code, kb_data_ptr)
+
+            mouse_pointer = create_pointer(mouse_low_level_handler);
+            self.id = windll.user32.SetWindowsHookExA(0x0E, mouse_pointer, windll.kernel32.GetModuleHandleW(None), 0)
+
+        atexit.register(windll.user32.UnhookWindowsHookEx, self.id)
+        while True:
+            msg = windll.user32.GetMessageW(None, 0, 0, 0)
+            windll.user32.TranslateMessage(byref(msg))
+            windll.user32.DispatchMessageW(byref(msg))
```
vasily-v-ryabov: It's better to move this code to separate method `listen()`.

pywinauto/hooks.py
```diff
@@ -0,0 +1,231 @@
...
+
+
+def on_event(args):
+    if isinstance(args, KeyboardEvent):
+        if args.current_key == 'A' and args.event_type == 'key down' and args.pressed_key.__contains__('Lcontrol'):
```
vasily-v-ryabov: Should be `'Lcontrol' in args.pressed_key` which is equivalent.

pywinauto/hooks.py
```diff
@@ -0,0 +1,231 @@
...
+
+            keyboard_pointer = create_pointer(low_level_handler);
+            windll.kernel32.GetModuleHandleW.restype = wintypes.HMODULE
+            windll.kernel32.GetModuleHandleW.argtypes = [wintypes.LPCWSTR]
+            self.id = windll.user32.SetWindowsHookExA(0x00D, keyboard_pointer, windll.kernel32.GetModuleHandleW(None),
+                                                      0)
+
+        if mouse:
+            def mouse_low_level_handler(code, event_code, kb_data_ptr):
+                if event_code != 512:
```
vasily-v-ryabov: What does this magic number `512` mean?

pywinauto/hooks.py
MaxSamokhvalov: It is very weird, but when I called this code in separate method (for example wait()) script was crashes. I spent a lot of time to fix it, but did not understand what the problem is.
`atexit.register(windll.user32.UnhookWindowsHookEx, self.id)
+        while True:
+            msg = windll.user32.GetMessageW(None, 0, 0, 0)
+            windll.user32.TranslateMessage(byref(msg))
+            windll.user32.DispatchMessageW(byref(msg))`

pywinauto/hooks.py
```diff
@@ -0,0 +1,231 @@
...
+    MouseId2EventType = {513: 'key down',
+                         514: 'key up',
+                         516: 'key down',
+                         517: 'key up',
+                         519: 'key down',
+                         520: 'key up',
+                         522: None}
+
+    ID2Key = {8: 'Back',
+              9: 'Tab',
```
vasily-v-ryabov: Function `SendKeysCtypes.parse_keys(...)` might be re-used to parse a key sequence. Returned objects can help to get key codes.

pywinauto/hooks.py
```diff
@@ -0,0 +1,231 @@
...
+                    event = MouseEvent(current_key, event_code)
+                    if self.handler != 0:
+                        self.handler(event);
+
+                return windll.user32.CallNextHookEx(self.id, code, event_code, kb_data_ptr)
+
+            mouse_pointer = create_pointer(mouse_low_level_handler);
+            self.id = windll.user32.SetWindowsHookExA(0x0E, mouse_pointer, windll.kernel32.GetModuleHandleW(None), 0)
+
+        atexit.register(windll.user32.UnhookWindowsHookEx, self.id)
```
vasily-v-ryabov: This is great for a peace of mind. But a user should have an ability to unhook the handler at the moment he wants. So if we have `hook()` method, why not have `unhook()` too?

pywinauto/hooks.py
```diff
@@ -4,41 +4,45 @@
 
 
 def create_pointer(handler):
+	"""Create and return C-pointer"""
```
vasily-v-ryabov: Seems like different indentation. Maybe forgot to set up `Replace tabs with spaces` in your editor?

pywinauto/hooks.py
```diff
@@ -0,0 +1,231 @@
...
+import atexit
+
+
+def create_pointer(handler):
+    cmp_func = CFUNCTYPE(c_int, c_int, c_int, POINTER(c_void_p))
+    return cmp_func(handler)
+
+
+class KeyboardEvent:
+    def __init__(self, current_key=None, event_type=None, pressed_key=[]):
```
airelil: Didn't have too much time to review the whole code. Just my two cents. Don't use list as a default argument. Use
```python
def append_to(element, to=None):
    if to is None:
        to = []
    to.append(element)
    return to
```
More details here: http://docs.python-guide.org/en/latest/writing/gotchas/

pywinauto/hooks.py
```diff
@@ -0,0 +1,242 @@
...
+                    event_code = self.MOUSE_ID_TO_EVENT_TYPE[event_code]
+
+                    event = MouseEvent(current_key, event_code)
+                    if self.handler != 0:
+                        self.handler(event);
+
+                return windll.user32.CallNextHookEx(self.id, code, event_code, kb_data_ptr)
+
+            mouse_pointer = create_pointer(mouse_low_level_handler);
+            self.id = windll.user32.SetWindowsHookExA(0x0E, mouse_pointer, windll.kernel32.GetModuleHandleW(None), 0)
```
airelil: @vasily-v-ryabov already wrote about the 'magic numbers' in the issue #43 itself. WH_KEYBOARD_LL/WH_MOUSE_L constants should be used instead

pywinauto/hooks.py
```diff
@@ -0,0 +1,242 @@
...
+				"""Execute when mouse low level event was catched"""
+                if event_code != 512:
+                    current_key = self.MOUSE_ID_TO_KEY[event_code]
+                    event_code = self.MOUSE_ID_TO_EVENT_TYPE[event_code]
+
+                    event = MouseEvent(current_key, event_code)
+                    if self.handler != 0:
+                        self.handler(event);
+
+                return windll.user32.CallNextHookEx(self.id, code, event_code, kb_data_ptr)
```
airelil: ```SetWindowsHookExA``` is used for the hook registration but ```CallNextHookEx``` and ```UnhookWindowsHookEx``` for the rest. Should they all be ansi versions: ```SetWindowsHookExA, CallNextHookExA, UnhookWindowsHookExA``` ? 
Or maybe simply: ```SetWindowsHookEx, CallNextHookEx, UnhookWindowsHookEx```

airelil:
Any ideas how to test the feature ? On what OS versions it's been checked ?

vasily-v-ryabov:
I have one idea how to test it. We may run separate script that hooks and listens for events and then run some test application and emulate some input. The `hook-and-listen.py` script can play 2 roles: a part of unit tests and an example of `hooks.py` usage (placed in `examples` folder).

pywinauto/hooks.py
```diff
@@ -0,0 +1,242 @@
...
+
+                if event_code == 'key up':
+                    self.pressed_keys.remove(current_key)
+
+                event = KeyboardEvent(current_key, event_code, self.pressed_keys)
+
+                if self.handler != 0:
+                    self.handler(event);
+
+                return windll.user32.CallNextHookEx(self.id, code, event_code, kb_data_ptr)
```
airelil: Looks as the wrong ```event_code``` passed through. Isn't it replaced with ``` event_code = self.event_types[event_code];``` ? (BTW, you don't need ```;``` at the end) The original parameter should be kept a side.

pywinauto/hooks.py
MaxSamokhvalov: When I added `from . import SendKeysCtypes as SendKeys` and launched it, i get exception: `SystemError: Parent module '' not loaded, cannot perform relative import`
Do you have any ideas why it happened?

pywinauto/hooks.py
MaxSamokhvalov: Fix it.

pywinauto/hooks.py
MaxSamokhvalov: Fixed.

pywinauto/hooks.py
MaxSamokhvalov: Fixed.

pywinauto/hooks.py
MaxSamokhvalov: Fixed.

pywinauto/hooks.py
MaxSamokhvalov: Unhook methods was added.

pywinauto/hooks.py
MaxSamokhvalov: Yes. Fixed.

pywinauto/hooks.py
MaxSamokhvalov: Fixed.

pywinauto/hooks.py
MaxSamokhvalov: Fixed.

pywinauto/hooks.py
MaxSamokhvalov: No. In all other case I will get crashes.

pywinauto/hooks.py
MaxSamokhvalov: Fixed.

MaxSamokhvalov:
@vasily-v-ryabov i have 2 problems:
1. I can't separate method `hook` on `hook_keyboard`, `hook_mouse` and `listen because` if i do it i get crashes (without exception). 
2. I created `hook-and-listen.py`, but when I try launch i get Exception
```
File "D:\Python\pywinauto\pywinauto\controls\HwndWrapper.py", line 34, in <module>
    import win32api
ImportError: No module named 'win32api'
```
Can you help me please with it?

All other problems iI fixed.

pywinauto/hooks.py
```diff
@@ -0,0 +1,276 @@
...
+        if args.current_key == 'WButton' and args.event_type == 'key down':
+            print("Wheel button pressed")
+
+        if args.current_key == 'WheelButton' and args.event_type == 'key down':
+            print("Wheel button pressed")
+
+
+hk = Hook()
+hk.handler = on_event
+hk.hook(keyboard=True, mouse=True)
```
airelil: Should be placed  under ```if __name__ == "__main__":```

pywinauto/hooks.py
```diff
@@ -0,0 +1,276 @@
...
+        """Hook mouse and/or keyboard events"""
+        if not mouse and not keyboard:
+            return
+
+        if keyboard:
+            def keyboard_low_level_handler(code, event_code, kb_data_ptr):
+                """Execute when keyboard low level event was catched"""
+                key_code = kb_data_ptr[0]
+                current_key = self.ID_TO_KEY[key_code]
+                event_type = self.event_types[event_code]
```
airelil: Generally, any input from the outer world can't be trusted so any of these dictionaries translations may fail. Not sure if we have a time for an extensive validation here so I'd suggest to guard with ```try``` and use ```finally``` to make sure we pass the event to the next-in-chain hook even if we failed to handle it.
The same treatment needed for ```mouse_low_level_handler```

pywinauto/hooks.py
```diff
@@ -0,0 +1,276 @@
...
+                if event_type == 'key down':
+                    self.pressed_keys.append(current_key)
+
+                if event_type == 'key up':
+                    self.pressed_keys.remove(current_key)
+
+                event = KeyboardEvent(current_key, event_type, self.pressed_keys)
+
+                if self.handler != 0:
+                    self.handler(event);
```
airelil: Same as above, the handler call has to be placed under a ```try``` guard

airelil:
I think you got ```ImportError: No module named 'win32api'``` because you didn't install pywin32

pywinauto/hooks.py
```diff
@@ -0,0 +1,276 @@
...
+
+    def unhook_keyboard(self):
+        """Unhook keyboard events"""
+        windll.user32.UnhookWindowsHookEx(self.keyboard_id)
+
+    def listen(self):
+        """Listen events"""
+        atexit.register(windll.user32.UnhookWindowsHookEx, self.keyboard_id)
+        atexit.register(windll.user32.UnhookWindowsHookEx, self.mouse_id)
+        while True:
```
airelil: The loop is infinite and it runs in the main thread. It should be stopped somehow at least after unhooking of all events

vasily-v-ryabov: `win32gui.Pu,pMessages()` also runs in infinite loop, but it exits when `WM_QUIT` happened in the application.

pywinauto/hooks.py
```diff
@@ -0,0 +1,276 @@
...
+            msg = windll.user32.GetMessageW(None, 0, 0, 0)
+            windll.user32.TranslateMessage(byref(msg))
+            windll.user32.DispatchMessageW(byref(msg))
```
vasily-v-ryabov: Did you try to set `restype` and `argtypes` for these functions? It can help to convert Python params to C params correctly.

vasily-v-ryabov:
[pyWin32 extensions](https://sourceforge.net/projects/pywin32/files/pywin32/) is a prerequisite for pywinauto. So you need to install them prior to unit tests run or some other debugging. I'm not sure we really need to add this dependency for `hooks.py` too.

pywinauto/hooks.py
airelil: @vasily-v-ryabov, I want to understand the use case of a such infinite loop. My understanding was a bit different. I supposed the hook should be activated at the beginning of a user script and help to interrupt the script when a predefined keys sequence has been pressed. This is a simple example I thought about
```python
hk = Hook()
hk.handler = fire_an_event_on_a_key_press('Ctrl-L') 
hk.hook(keyboard = True)

app = Application().start("gui app.exe")
dlg = app.Dialog
while True:
    dlg.do_an_extensieve_job...()
    if hook_event_occured:
        return
```
Using the hook will allow to stop the user script without bringing the Python console to foreground as it can be challenging when the script constantly interacts with GUI elements.

pywinauto/hooks.py
MaxSamokhvalov: Fixed.

pywinauto/hooks.py
MaxSamokhvalov: Try and finally was added.

pywinauto/hooks.py
MaxSamokhvalov: Fixed.

pywinauto/hooks.py
MaxSamokhvalov: I added condition for exit from infinite loop.

vasily-v-ryabov: It's up to user whether he wants to stop the script (say `exit(0)` by `Ctrl+C`) or continue capturing events. For example, if one wants to load some data into the application using GUI automation (by `Shift+1` for example), he may want to do that several times since automated tasks are often repeatable.

Another example to interrupt the script: http://stackoverflow.com/questions/6023172/ending-a-program-mid-run I'm not sure, but it seems `ctypes.windll.user32.PostQuitMessage(0)` does almost the same as `exit(0)`.

MaxSamokhvalov:
I intalled pyWin32 and hook-and-listen.py script to successfully work. Thanks.

examples/hook-and-listen.py
```diff
@@ -4,13 +4,22 @@
...
     if isinstance(args, KeyboardEvent):
         if args.current_key == 'A' and args.event_type == 'key down' and 'Lcontrol' in args.pressed_key:
             print("Ctrl + A was pressed");
 
         if args.current_key == 'K' and args.event_type == 'key down':
             print("K was pressed");
 
+        if args.current_key == 'M' and args.event_type == 'key down' and 'U' in args.pressed_key:
+            hk.unhook_mouse()
+            print("Unhook mouse")
```
vasily-v-ryabov: `from __future__ import print_function` is required at the beginning of the module (before all other imports) to make it working on both Py2 and Py3.

pywinauto/hooks.py
```diff
@@ -0,0 +1,286 @@
...
+        if not self.mouse_hook and not self.keyboard_hook:
+            return
+
+        if self.keyboard_hook:
+            def keyboard_low_level_handler(code, event_code, kb_data_ptr):
+                """Execute when keyboard low level event was catched"""
+                try:
+                    key_code = kb_data_ptr[0]
+                    current_key = self.ID_TO_KEY[key_code]
+                    event_type = self.event_types[event_code]
```
airelil: I would cleanup the params anyway
```python
key_code = 0xFFFFFFFF & kb_data_ptr[0]
current_key = self.ID_TO_KEY[key_code]
event_type = self.event_types[0xFFFFFFFF & event_code] 
```

pywinauto/hooks.py
MaxSamokhvalov: Sorry, I don't understand. Should I add  `exit(0)` in callback by `Ctrl+C`?

pywinauto/hooks.py
MaxSamokhvalov: Yes. It did not help.

pywinauto/hooks.py
vasily-v-ryabov: No, `Ctrl+C` is just for example. You don't need to implement it.

pywinauto/hooks.py
vasily-v-ryabov: OK, I'll take a look at this.

pywinauto/hooks.py
MaxSamokhvalov: Then what should I fix here?

vasily-v-ryabov: Here nope. It's just an explanation for @airelil.

vasily-v-ryabov:
Started manual testing of the feature...
This statement doesn't work on Python 2.7 (64-bit):
```python
            self.keyboard_id = windll.user32.SetWindowsHookExA(self.WH_KEYBOARD_LL, keyboard_pointer,
                                                               windll.kernel32.GetModuleHandleA(None),
                                                               0)
```

Traceback:
```
  File "C:\Python27_x64\Lib\site-packages\pythonwin\pywin\framework\scriptutils.py", line 326, in RunScript
    exec codeObject in __main__.__dict__
  File "C:\OneDrive\python_students\2016\04_Samokhvalov\pywinauto\test_hooks.py", line 14, in <module>
    hook.hook()
  File "C:\OneDrive\python_students\2016\04_Samokhvalov\pywinauto\pywinauto\hooks.py", line 214, in hook
    0)
ArgumentError: argument 3: <type 'exceptions.OverflowError'>: long int too long to convert
```
It looks like third argument must be coverted using `HINSTANCE` in `argtypes`. It's defined in `win32structures.py` so:
```python
if sysinfo.is_x64_Python():
    HINSTANCE = c_longlong
else:
    HINSTANCE = c_long
```
It can be re-defined in `hooks.py` to make it independent from the rest of pywinauto code.

vasily-v-ryabov:
I've just tried the following simple code:
```python
from pywinauto import hooks
hook = hooks.Hook()
hook.hook()
```

vasily-v-ryabov:
This is the illustration how it's converted:
```python
>>> int(ctypes.windll.kernel32.GetModuleHandleA(None))
5368709120L
>>> ctypes.c_long(5368709120L)
c_long(1073741824) # obviously incorrect
>>> ctypes.c_longlong(5368709120L)
c_longlong(5368709120L)
>>> 
```

pywinauto/hooks.py
MaxSamokhvalov: Fixed.

MaxSamokhvalov:
@vasily-v-ryabov I fixed it. But I have installed 64-bit verision Python 2.7 (for checking this bug), and all is work. I.e. I can't get this crashes with c_int.

examples/hook-and-listen.py
MaxSamokhvalov: Fixed.

airelil:
Hi All. I also tried to run the latest version of hooks.py (as in 4515774d4ea76c04572599b35d9918aff997fa5d) It works for me on my Anaconda versions: Py27 x64 and Py34 x64. However, I'm still questioning the ```listen``` method implementation and its use cases. See my inline comments about it.

pywinauto/hooks.py
```diff
@@ -0,0 +1,290 @@
...
+        """Unhook keyboard events"""
+        self.keyboard_hook = False
+        windll.user32.UnhookWindowsHookEx(self.keyboard_id)
+
+    def listen(self):
+        """Listen events"""
+        atexit.register(windll.user32.UnhookWindowsHookEx, self.keyboard_id)
+        atexit.register(windll.user32.UnhookWindowsHookEx, self.mouse_id)
+        while self.mouse_hook or self.keyboard_hook:
+            msg = windll.user32.GetMessageW(None, 0, 0, 0)
```
airelil: @MaxSamokhvalov, please look in [GetMessage function](https://msdn.microsoft.com/en-us/library/windows/desktop/ms644936(v=vs.85).aspx). This is completely wrong. You should provide a pointer to a message structure instead of ```None```. The return value is an error code not the message itself. The return value should be verified and only in case of success the ```TranslateMessage / DispatchMessageW``` should be called. I have to admit, I didn't get it from the beginning but the code not looping it just sits inside the ```GetMessageW```. What messages it waits for ? Only WM_QUIT ? Do we expect any other messages to this process ? @vasily-v-ryabov, I'd like to see a good example of using the hook. Is the hook supposed to run in a separate thread or even process ?

vasily-v-ryabov: In theory we need to wait for every message (`GetMessage` should return if some message is received as far as I understand). We just need to quit on `WM_QUIT`. But I didn't look into `PumpMessages()` implementation yet to know it for sure.

airelil: So maybe we need simply run ```pythoncom.PumpMessages()``` without re-inventing a wheel. Or even better, use the non-blocking ```pythoncom.PumpWaitingMessages()```

vasily-v-ryabov:
@MaxSamokhvalov, with [this fix](https://github.com/MaxSamokhvalov/pywinauto/pull/6) it passes `SetWindowsHookExA` at least. It's not so hard to make correct `argtypes` for the rest of functions. It's absolutely required.

Let's try to fix the issues without pyWin32 dependency. I think it's possible.

MaxSamokhvalov:
@vasily-v-ryabov you wrote:
```
Message loop still fails.

  File "C:\...\04_Samokhvalov\pywinauto\pywinauto\hooks.py", line 262, in listen
    msg = windll.user32.GetMessageW(None, 0, 0, 0)
WindowsError: exception: access violation reading 0x0000000000000008"
```
Where did you get this error? How can I catch it?
I'm just a little confused. What is the problem we solve?

vasily-v-ryabov:
I've got this error executing the following simple script on Python 2.7 x64:
```python
from pywinauto import hooks
hook = hooks.Hook()
hook.hook()
```
We definitely need to define an [MSG structure](https://msdn.microsoft.com/en-us/library/windows/desktop/ms644958(v=vs.85).aspx) for both 32-bit and 64-bit Pythons and pass this structure pointer to the `GetMessage` call. By the way, all retrieved fields might be added to the Python-side callback params.

MaxSamokhvalov:
@vasily-v-ryabov I think I fixed it. Plese, check it.

vasily-v-ryabov:
Still fails, but at the next statement:
```python
  File "C:\...\pywinauto\test_hooks.py", line 14, in <module>
    hook.hook()
  File "C:\...\pywinauto\pywinauto\hooks.py", line 258, in hook
    self.listen()
  File "C:\...\pywinauto\pywinauto\hooks.py", line 279, in listen
    windll.user32.TranslateMessage(byref(msg))
TypeError: byref() argument must be a ctypes instance, not 'int'
```

pywinauto/hooks.py
```diff
@@ -0,0 +1,312 @@
...
+
+if sizeof(POINTER(c_int)) * 8 == 64:
+    hinstance = c_longlong
+else:
+    hinstance = c_long
+
+cmp_func = CFUNCTYPE(c_int, c_int, hinstance, POINTER(c_void_p))
+DWORD = c_ulong
+
+windll.kernel32.GetModuleHandleA.restype = wintypes.HMODULE
```
vasily-v-ryabov: `wintypes.DWORD` and others is a great finding! It means we can eliminate a lot of win32 definitions from `win32defines.py` by using `wintypes`. I'm surprised how we missed it.

MaxSamokhvalov:
@vasily-v-ryabov and now?

vasily-v-ryabov:
Now it fails so
```python
  File "C:\...\pywinauto\pywinauto\hooks.py", line 280, in listen
    windll.user32.TranslateMessage(byref(msg))
TypeError: byref() argument must be a ctypes instance, not 'bool'
```
The problem is not in `GetMessage` but at the 2 next lines. Read the MSDN docs for every called `user32.dll` function.

MaxSamokhvalov:
@vasily-v-ryabov should I set MSG pointer in arstypes TranslateMessage и DispatchMessageW?

So?
windll.user32.TranslateMessage.argtypes = [POINTER(MSG)]
windll.user32.DispatchMessageW.argtypes = [POINTER(MSG)]

vasily-v-ryabov:
@MaxSamokhvalov, correct. And pass `message_pointer` instead of `msg`, because these functions require `MSG` structure instance, not an integer message id returned by previous call.

MaxSamokhvalov:
@vasily-v-ryabov check it, please.
Blind programming :)

vasily-v-ryabov:
@MaxSamokhvalov, OK, now it doesn't crash. But we need to exit on `WM_QUIT` message (i.e. break a loop `if message_pointer.message == WM_QUIT`). `WM_QUIT = 0x0012` from the MSDN.

airelil:
@MaxSamokhvalov I also don't get any messages when I run the hooks.py. So basically, all the loop is freezed. @vasily-v-ryabov how do you run your Python interpreter so that you get windows messages ? Do you use some IDE ? I use a small script to send WM_QUIT to myself. It's based on your Stackoverflow [link](http://stackoverflow.com/questions/6023172/ending-a-program-mid-run)
@MaxSamokhvalov use that snippet to debug your code
```python
import win32api
import win32con
from threading import Timer
import hooks

def on_timer():
    win32api.PostThreadMessage(main_thread_id, win32con.WM_QUIT, 0, 0);

def on_event(args):
    """Callback for keyboard and mouse events"""
    if isinstance(args, KeyboardEvent):
        hk.unhook_keyboard()
        print("Unhook keyboard")

hk = hooks.Hook()
hk.handler = on_event
main_thread_id = win32api.GetCurrentThreadId()
t = Timer(5.0, on_timer) # Quit after 5 seconds
t.start()
hk.hook(keyboard=True, mouse=False)

```

pywinauto/hooks.py
```diff
@@ -0,0 +1,317 @@
...
+
+    def listen(self):
+        """Listen events"""
+        atexit.register(windll.user32.UnhookWindowsHookEx, self.keyboard_id)
+        atexit.register(windll.user32.UnhookWindowsHookEx, self.mouse_id)
+
+        message_pointer = MSG()
+
+        while self.mouse_is_hook or self.keyboard_is_hook:
+            msg = windll.user32.GetMessageW(message_pointer, 0, 0, 0)
```
airelil: @MaxSamokhvalov you don't check the return value. Did you read the MSDN about [GetMessage](https://msdn.microsoft.com/en-us/library/windows/desktop/ms644936(v=vs.85).aspx) that I pointed out ? I'd still consider to use non-blocking functions. For example, using [PeekMessage](https://msdn.microsoft.com/en-us/library/windows/desktop/ms644943(v=vs.85).aspx) will allow to be more responsive with more frequent checking the main condition ```while self.mouse_is_hook or self.keyboard_is_hook```

pywinauto/hooks.py
```diff
@@ -0,0 +1,317 @@
...
+DWORD = c_ulong
+
+windll.kernel32.GetModuleHandleA.restype = wintypes.HMODULE
+windll.kernel32.GetModuleHandleA.argtypes = [wintypes.LPCWSTR]
+
+windll.user32.SetWindowsHookExA.restype = c_int
+windll.user32.SetWindowsHookExA.argtypes = [c_int, cmp_func, hinstance, DWORD]
+
+
+class MSG(Structure):
```
airelil: If you already discovered ```wintypes``` and started using it, why duplicate the ```MSG``` definition? Just because of ```hinstance```? It's also should be dropped in favor of ```wintypes.HANDLE```

MaxSamokhvalov:
@airelil i launched you snippet - it work (I got message `Unhook keyboard`).
There only thing - i got error when snippet exit by timer:
```
File "C:\Python\pywinauto\pywinauto\hooks.py", line 280, in listen
    msg = windll.user32.GetMessageW(message_pointer, 0, 0, 0)
WindowsError: exception: access violation reading 0x00000004

Process finished with exit code 1
```
P.S. My IDE is PyCharm CE 2016.1.2

airelil:
@MaxSamokhvalov , this is exactly the point. Now you can debug the problems in the code.

MaxSamokhvalov:
@airelil I made a mistake - I get exception only on 32-bit version Python (without reference to I use GetMessage or PeekMessage).
On 64-bit version this snippet works fine without exception.

airelil:
@MaxSamokhvalov the code crashes on my Python x86 too. I think it's because of wrong parameters to ```GetMessage ```. MSG must be passed as a pointer. ```message_pointer = MSG()``` is just an object.
Change the argtypes to get the pointer on the message. The following works on x86 and x64 for me:
```python
windll.user32.GetMessageW.argtypes = [POINTER(MSG), hinstance, c_uint, c_uint]
...
message = MSG()
...
windll.user32.GetMessageW(byref(message), 0, 0, 0)
```
When you get fixed the ```GetMessageW``` call you need to make more testing. Try to change my snippet. Prepare and send a message different from  ```WM_QUIT```. Read MSDN about GetMessage/PeekMessage. Verify return values and act according to it.

MaxSamokhvalov:
@airelil Thank you.
I fixed crashes.
I added handler for error (in this case GetMessage return -1 according to documentation).
Also I changed checking on WM_QUIT (in this case GetMessage always return 0).

pywinauto/hooks.py
```diff
@@ -262,11 +261,17 @@ def mouse_low_level_handler(code, event_code, kb_data_ptr):
 
     def unhook_mouse(self):
         """Unhook mouse events"""
+        if not self.mouse_is_hook:
```
vasily-v-ryabov: `if self.mouse_is_hook:` and then 2 lines is shorter. Empty `return` is not needed then.

pywinauto/hooks.py
```diff
@@ -275,14 +280,22 @@ def listen(self):
...
+        message = MSG()
 
         while self.mouse_is_hook or self.keyboard_is_hook:
-            msg = windll.user32.GetMessageW(message_pointer, 0, 0, 0)
-            windll.user32.TranslateMessage(byref(message_pointer))
-            windll.user32.DispatchMessageW(byref(message_pointer))
-            if message_pointer.message == self.WM_QUIT:
+            msg = windll.user32.GetMessageW(byref(message), 0, 0, 0)
+            if msg == -1:
+                print("Error. Sorry, I exit... :(")
```
vasily-v-ryabov: Debug output must be removed from production code.

pywinauto/hooks.py
```diff
@@ -275,14 +280,22 @@ def listen(self):
...
+                hk.unhook_keyboard()
+                hk.unhook_mouse()
+                exit(0)
```
vasily-v-ryabov: I'm not sure `exit(0)` is what we need. Just break the loop, do not exit the whole Python process.

pywinauto/hooks.py
MaxSamokhvalov: Fixed.

pywinauto/hooks.py
MaxSamokhvalov: Fixed.

pywinauto/hooks.py
airelil: So what about this MSG duplicate? I still think we need to switch to ```wintypes.MSG``` and ```wintypes.HANDLE``` instead of the custom ```hinstance```

airelil:
hook-and-listen.py is not PEP8 name. Should be renamed to hook_and_listen.py

pywinauto/hooks.py
MaxSamokhvalov: If I use wintypes.MSG, Vasiliy get an error, isnt it?

airelil: If it fails on any computer it means that there is an additional bug. Like previous one with x64/x86 when it crashed on x86 only. And I wouldn't blame ```ctypes``` on it before checking my own code.

vasily-v-ryabov: I didn't try `wintypes.MSG` but agree with @airelil: we need to re-use maximum of `wintypes` functionality since it's so rich and built-in (i.e. no 3rd party dependency).

MaxSamokhvalov:
I renamed hook_and_listen.py

pywinauto/hooks.py
MaxSamokhvalov: I replaced custom MSG with MSG from wintypes (also replaced some types)

vasily-v-ryabov:
OK, now it seems working well and we have interesting findings. @MaxSamokhvalov, thank you for the good job done.

MaxSamokhvalov:
@vasily-v-ryabov @airelil thank you for new features that I have learned.

PR #195, opened 2016-05-04
Egor Sharov

EgorSharov:
task N30


pywinauto/unittests/test_Calendar.py
```diff
@@ -92,7 +92,24 @@ def assert_system_time_is_equal_to_current_date_time(self,systemTime, now):
...
+        self.app['Common Controls Sample']['MCS_MULTISELECT'].WrapperObject().Click()
+
+    def test_can_get_today(self):
+        date = self.calendar.get_today()
+        self.assert_system_time_is_equal_to_current_date_time(date, datetime.date.today())
+
+    def test_can_set_today(self):
+        self.calendar.set_today(2016, 5, 1)
+        self.assert_system_time_is_equal_to_current_date_time(self.calendar.get_today(),
+                                                              datetime.date(2016, 5, 1))
```
airelil: Document all the test methods with three-quote doc strings: ``` """Test setting up the control's today field"""```

pywinauto/controls/common_controls.py
```diff
@@ -3388,7 +3388,58 @@ def set_view(self, viewType):
...
+                system_date = win32structures.SYSTEMTIME()
+                remote_mem.Write(system_date)
+
+                res = self.send_message(win32defines.MCM_GETTODAY, 0, remote_mem)
+                remote_mem.Read(system_date)
+                del remote_mem
+
+                if res == 0:
+                    raise RuntimeError('Failed to get today date in Calendar')
+                return system_date
```
airelil: Bad indentation

EgorSharov:
@airelil Thanks for comments.

pywinauto/unittests/test_Calendar.py
vasily-v-ryabov: Landscape.io is not checking `test_` method docstrings. So I would say it's absolutely required for test methods.

pywinauto/unittests/test_Calendar.py
```diff
@@ -92,7 +92,28 @@ def assert_system_time_is_equal_to_current_date_time(self,systemTime, now):
...
+
+    def test_can_set_today(self):
+        """Test setting up the control's today field"""
+        self.calendar.set_today(2016, 5, 1)
+        self.assert_system_time_is_equal_to_current_date_time(self.calendar.get_today(),
+                                                              datetime.date(2016, 5, 1))
+
+    def test_is_first_day_of_week_locale(self):
+        """Test is the today field set as LOCALE_IFIRSTDAYOFWEEK"""
+        self.assertEqual(0, self.calendar.get_first_weekday())
```
vasily-v-ryabov: This test would fail on many locales. It's easier to leave just `set/get` test because it doesn't depend on the OS settings.

pywinauto/unittests/test_Calendar.py
```diff
@@ -92,7 +92,28 @@ def assert_system_time_is_equal_to_current_date_time(self,systemTime, now):
...
+
+    def test_can_set_and_get_first_day_of_week(self):
+        """Test can set and get first day of the week"""
+        self.calendar.set_first_weekday(4)
+        self.assertEqual(4, self.calendar.get_first_weekday()-65536)
```
vasily-v-ryabov: Why this `-65536` is required here? It looks wierd. We set `4` but `get_` method returns another value. Maybe you've missed `int -> wintypes.DWORD` conversion inside the method?

EgorSharov: @vasily-v-ryabov Can you give me advice ? 
MCM_GETFIRSTDAYOFWEEK Returns a DWORD value that contains two values. The high word is a BOOL value that is nonzero if the first day of the week is set to something other than LOCALE_IFIRSTDAYOFWEEK, or zero otherwise. The low word is an INT value that represents the first day of the week, where 0 is Monday, 1 is Tuesday, and so on.

What should I return? Just the low word from 0 to 6 ?

vasily-v-ryabov: Module `win32functions.py` contains functions `HiWord` and `LoWord` for that purpose. You can return a tuple: `(is_in_current_locale, first_day_of_week)` where first one is a `bool` and second is an `int`.

EgorSharov: @vasily-v-ryabov  Thanks for fast reply, I've changed logic according your remarks.

pywinauto/unittests/test_Calendar.py
```diff
@@ -92,7 +92,31 @@ def assert_system_time_is_equal_to_current_date_time(self,systemTime, now):
...
+    def test_can_set_and_get_first_day_of_week(self):
+        """Test can set and get first day of the week"""
+        self.calendar.set_first_weekday(4)
+        self.calendar.get_first_weekday()
+        self.assertEqual(4, self.calendar.get_first_weekday()[1])
+
+    def test_is_not_locale_when_setting_first_day_of_the_week(self):
+        """Test is locale changes to 1 when the first day of the week other than LOCALE_IFIRSTDAYOFWEEK,"""
+        self.calendar.set_first_weekday(4)
+        self.calendar.get_first_weekday()
```
vasily-v-ryabov: Why this return value is not checked?

pywinauto/unittests/test_Calendar.py
```diff
@@ -92,7 +92,31 @@ def assert_system_time_is_equal_to_current_date_time(self,systemTime, now):
...
+    def test_can_set_today(self):
+        """Test setting up the control's today field"""
+        self.calendar.set_today(2016, 5, 1)
+        self.assert_system_time_is_equal_to_current_date_time(self.calendar.get_today(),
+                                                              datetime.date(2016, 5, 1))
+
+    def test_can_set_and_get_first_day_of_week(self):
+        """Test can set and get first day of the week"""
+        self.calendar.set_first_weekday(4)
+        self.calendar.get_first_weekday()
```
vasily-v-ryabov: The same as below. No assert here.

EgorSharov: Sorry, forget to commit changes! It's useless line.

pywinauto/unittests/test_Calendar.py
```diff
@@ -92,7 +92,29 @@ def assert_system_time_is_equal_to_current_date_time(self,systemTime, now):
...
+
+    def test_can_set_and_get_first_day_of_week(self):
+        """Test can set and get first day of the week"""
+        self.calendar.set_first_weekday(4)
+        self.assertEqual(4, self.calendar.get_first_weekday()[1])
+
+    def test_is_not_locale_when_setting_first_day_of_the_week(self):
+        """Test is locale changes to 1 when the first day of the week other than LOCALE_IFIRSTDAYOFWEEK,"""
+        self.calendar.set_first_weekday(4)
+        self.assertEqual(1, self.calendar.get_first_weekday()[0])
```
vasily-v-ryabov: OK, why not check the whole tuple in both cases?

EgorSharov: Ok, changed  in current commit

pywinauto/unittests/test_Calendar.py
```diff
@@ -92,7 +92,24 @@ def assert_system_time_is_equal_to_current_date_time(self,systemTime, now):
...
+        """Test setting up the control's today field"""
+        self.calendar.set_today(2016, 5, 1)
+        self.assert_system_time_is_equal_to_current_date_time(self.calendar.get_today(),
+                                                              datetime.date(2016, 5, 1))
+
+    def test_can_set_and_get_first_day_of_week(self):
+        """Test can set and get first day of the week"""
+        self.calendar.set_first_weekday(4)
+        self.assertEqual(4, self.calendar.get_first_weekday()[1])
+        self.assertEqual(1, self.calendar.get_first_weekday()[0])
```
vasily-v-ryabov: `self.assertEqual((True, 4), self.calendar.get_first_weekday())` should pass in a final version.

EgorSharov: Thanks for advice, that passed!

vasily-v-ryabov:
OK, you're lucky today. :)

PR #196, opened 2016-05-05
Changing calc.exe for another apps in unit tests

nikkrasn:
I changed some tests, replacing calc.exe for another apps in MFC_Samples folder. Need some help with testWriteToXML test. It works fine while it was a calc.exe app, but since i changed it, i'm getting 'Key error' now. How i can fix this?


pywinauto/unittests/test_win32controls.py
```diff
@@ -54,110 +54,87 @@
...
+        """Test getting the properties for the button control"""
+        props = self.ctrl.Button2.GetProperties()
 
         self.assertEquals(
-            "RadioButton", props['friendly_class_name'])
+            "Button", props['friendly_class_name'])
 
         self.assertEquals(
-            self.calc.Degrees.texts(), ['Degrees'])
+            self.ctrl.Button2.texts(), ['Отмена'])
```
vasily-v-ryabov: The tests must work on Windows with English or any other interface language. Please choose another control with English-only names.

pywinauto/unittests/test_win32controls.py
```diff
@@ -591,69 +568,64 @@ def testSetEditTextWithUnicode(self):
         #self.ctrl.Select()
         #self.ctrl.SetEditText(u'\u043d\u0435\u0447\u0442', pos_end=2) # u'necht'
         #self.assertEquals(self.ctrl.TextBlock(), u'\u043d\u0435\u0447\u0442\u043e') # u'nechto'
-
+#
 
 class DialogTestCases(unittest.TestCase):
-
```
vasily-v-ryabov: Empty line is required before class docstring. Do not remove them please. This would produce Landscape.io warnings.

pywinauto/unittests/test_UIAElementInfo.py
```diff
@@ -16,16 +22,12 @@ def setUp(self):
...
             self.app = Application(backend="native")
-            if is_x64_Python() or not is_x64_OS():
-                self.app.start(r"C:\Windows\System32\calc.exe")
-            else:
-                self.app.start(r"C:\Windows\SysWOW64\calc.exe")
-            
-            self.dlg = self.app.Calculator
+            self.app = self.app.Start(mfc_app_1)
+
+            self.dlg = self.app.RowListSampleApplication
```
vasily-v-ryabov: It's better to use an app from `WPF_samples` folder for UIA tests.

nikkrasn:
Make some changes, but still having problems with testWriteToXML.

nikkrasn:
I think i just realized that's the problem
When this code is processing
all_props.extend([c.GetProperties() for c in self.cmn_ctrl.children()])
It extend props with only default properties, but TreeView in CmnCtrl1.exe have an additional property called 'item_count'
So when it tries to find property by the key 'item_count' i get this error
for key, ctrl_value in ctrl.items():
>               expected_value = all_props[i][key]
E               KeyError: 'item_count'
Any thoughts how can i avoid this?

vasily-v-ryabov:
It looks like a bug in `WriteDialogToFile` function (`XMLHelpers.py`). Thanks for finding this. But I cannot merge PR with failed tests. Could you try to take a look into `WriteDialogToFile` with some debugging? Maybe the fix is not so complicated.

vasily-v-ryabov:
Pitfalls are possible and you might rely on that. I'm finishing reviews for today. May have a short look in the morning, but not sure I will have a time to review all the fixes. There are too many students who've tried to have done their tasks in the last moment.

nikkrasn:
I just found a solution for my problem. Then this code is processing:
`all_props.extend([c.GetProperties() for c in self.cmn_ctrl.children()])`
It calls a function `get_properties` from `base_wrapper.py`, where we have this code
```
# for each of the properties that can be written out
        for propname in self.writable_props:
            # set the item in the props dictionary keyed on the propname
            props[propname] = getattr(self, propname)()
```
For some reason `writable_props` for `TreeView` and `TabControl` are not extended. So `item_count` for `TreeView` and `tab_count` for `TabControl` simply not get at all.
After I added extension to `writable_props` for `TreeViewWrapper` and `TabControlWrapper` in `common_controls.py` this code worked fine for me.

vasily-v-ryabov:
Brilliant! Thanks for the good job done.

PR #193, opened 2016-05-04
print_administrators.py example script

skachkova-natalia:
Added computer_administrators_script.py

PR #198, opened 2016-05-05
Implement a timeout for connect() from application

SerovAlexey:


SerovAlexey:
@vasily-v-ryabov , check, please

pywinauto/application.py
```diff
@@ -839,6 +840,15 @@ def __connect(self, **kwargs):
...
         """Connects to an already running process"""
         connected = False
+
+        if 'timeout' in kwargs:
+            try:
+                self.timeout = float(kwargs['timeout'])
+                time.sleep(self.timeout)
+            except Exception as e:
+                message = "timeout isn't correct"
+                raise AppNotConnected(str(e) + message)
```
airelil: The exception should be ```ValueError``` for the wrong argument. Also using ```time.sleep``` is a naive approach. The method should try to check it several times during the timeout period and continue as soon as the app successfully connected. You might try to use ```WaitUntilPasses``` with a list of possible exceptions for that purpose or even
reuse an idea with a decorator that I suggested in another PR: https://github.com/pywinauto/pywinauto/pull/193#discussion_r62149644

SerovAlexey:
@airelil @vasily-v-ryabov fived

vasily-v-ryabov:
Local test run fails (with `C:\Python27_x64\python.exe .\pywinauto\unittests\test_application.py`):
```
ERROR: test_connect_with_App (__main__.ApplicationTestCases)
----------------------------------------------------------------------
  File ".\pywinauto\unittests\test_application.py", line 609, in test_connect_with_App
    self.app.Connect(path=path, timeout=1)
  File ".\pywinauto\application.py", line 851, in connect
    raise ValueError(str(e))
ValueError: Could not find any accessible process with a module of 'c:\windows\system32\notepad.exe'

----------------------------------------------------------------------
Ran 42 tests in 98.103s

FAILED (errors=1)
```

pywinauto/application.py
```diff
@@ -839,6 +841,15 @@ def __connect(self, **kwargs):
     def connect(self, **kwargs):
         """Connects to an already running process"""
         connected = False
+
+        if 'timeout' in kwargs:
+            try:
+                self.timeout = float(kwargs['timeout'])
+                del kwargs['timeout']
+                WaitUntilPasses(self.timeout, Timings.app_connect_retry, partial(self.Connect, **kwargs), (ValueError))
```
airelil: This is an interesting trick with ```partial```. Use PEP-8 name for the method though : ```self.connect```

pywinauto/unittests/test_application.py
```diff
@@ -566,6 +565,51 @@ def test_kill_(self):
...
+        """Test that connect_() works with a timeout"""
+        app1 = Application()
+        app1.start(_notepad_exe())
+
+        app_conn = Application()
+        if is_x64_Python() or not is_x64_OS():
+            app_conn.connect(path = r"c:\windows\system32\notepad.exe", timeout = "1.0")
+
+        else:
+            app_conn.connect(path = r"c:\windows\syswow64\notepad.exe", timeout = "1.0")
```
airelil: You can't use absolute paths. You'd better switch to our samples apps

pywinauto/unittests/test_application.py
```diff
@@ -566,6 +565,51 @@ def test_kill_(self):
...
+
+        app_conn.UntitledNotepad.MenuSelect('File->Exit')
```
airelil: You should use unittest module assert methods to verify results of your tests.

pywinauto/timings.py
```diff
@@ -111,6 +111,9 @@ class TimeConfig(object):
         'app_start_timeout' : 10.,
         'app_start_retry' : .90,
 
+        'app_connect_timeout': 5.,
```
airelil: It's not used. Please remove

vasily-v-ryabov: Maybe it's better to use this value? In case when `timeout=None` by default. It could allow us manipulating it as all other timings.

pywinauto/unittests/test_application.py
vasily-v-ryabov: Yes, `notepad.exe` is a real world application that is not intended to be used everywhere in the unit tests.

pywinauto/application.py
```diff
@@ -839,6 +841,15 @@ def __connect(self, **kwargs):
     def connect(self, **kwargs):
         """Connects to an already running process"""
         connected = False
+
+        if 'timeout' in kwargs:
+            try:
+                self.timeout = float(kwargs['timeout'])
```
vasily-v-ryabov: We don't need to save `timeout` as an attribute of `self` because we don't need it in any other methods.

pywinauto/application.py
vasily-v-ryabov: And `wait_until_passes` too.

SerovAlexey: You may suggest how to do it? I have not found how to avoid "connect () takes one positional argument but two were given"

SerovAlexey:
@vasily-v-ryabov @airelil  check, please

pywinauto/unittests/test_application.py
```diff
@@ -602,12 +606,9 @@ def test_connect_with_App(self):
...
 
         self.app = Application().start(os.path.join(mfc_samples_folder, u"CmnCtrl1.exe"))
-        if is_x64_Python() or not is_x64_OS():
-            path=r"c:\windows\system32\notepad.exe"
-        else:
-            path=r"c:\windows\syswow64\notepad.exe"
-        self.app.Connect(path=path, timeout=1)
-        self.app.start(path)
+        self.app1 = Application().start(os.path.join(mfc_samples_folder, u"CmnCtrl2.exe"))
+        self.app.Connect(path=os.path.join(mfc_samples_folder, u"CmnCtrl2.exe"), timeout=10)
```
vasily-v-ryabov: Use PEP8 name `connect()` instead of `Connect()`.

pywinauto/application.py
```diff
@@ -844,9 +844,10 @@ def connect(self, **kwargs):
 
         if 'timeout' in kwargs:
             try:
-                self.timeout = float(kwargs['timeout'])
+                timeout = float(kwargs['timeout'])
                 del kwargs['timeout']
                 WaitUntilPasses(self.timeout, Timings.app_connect_retry, partial(self.Connect, **kwargs), (ValueError))
+                #WaitUntilPasses(timeout, Timings.app_connect_retry, self.connect, (ValueError), kwargs)
```
vasily-v-ryabov: Use PEP8 name `wait_until_passes` instead of `WaitUntilPasses`.

vasily-v-ryabov:
1 test is failing. Deep into the `Details` link at the `AppVeyor build filed`, dive into any red line and scroll down until error traceback in the output. Or check the unit tests locally (`python .\pywinauto\unittests\test_application.py` runs much faster than `testall.py`).
app_veyor_tests

SerovAlexey:
@vasily-v-ryabov fixed

pywinauto/application.py
```diff
@@ -839,6 +841,15 @@ def __connect(self, **kwargs):
     def connect(self, **kwargs):
         """Connects to an already running process"""
         connected = False
+
+        if 'timeout' in kwargs:
+            try:
+                timeout = float(kwargs['timeout'])
+                del kwargs['timeout']
+                wait_until_passes(self.timeout, Timings.app_connect_retry, partial(self.Connect, **kwargs), (ValueError))
```
airelil: I don't understand it. ```timeout``` argument is stored into a local variable and deleted from ```kwargs``` but the actual call ```wait_until_passes``` still uses ```self.timeout``` that is only initiated to ```0``` and never changed ? How it supposed to work. Where is a test to check that the actual timeout period has been expired and the exception has been raised ? Am I missing something ?

vasily-v-ryabov:
@SerovAlexey please run & fix the tests locally first and commit only after that.

SerovAlexey:
@vasily-v-ryabov  I'm doing it

pywinauto/application.py
SerovAlexey: @airelil typo, instead of self.timeout to be timeout - it's fixed.

SerovAlexey:
@vasily-v-ryabov ,check, please,
changed the method Connect () method because wait_until passes () you proposed doing a not something that need

pywinauto/application.py
```diff
@@ -846,7 +845,7 @@ def connect(self, **kwargs):
             try:
                 timeout = float(kwargs['timeout'])
                 del kwargs['timeout']
-                wait_until_passes(self.timeout, Timings.app_connect_retry, partial(self.Connect, **kwargs), (ValueError))
+                time.sleep(timeout)
```
vasily-v-ryabov: Hey, what if an app can start from 5 to 30 seconds and you wait always for 35 seconds. It's very hard-coded way. You need to wait flexibly for better performance on fast machines. That why we have this task.

`time.sleep` can be called outside without any additional tricks inside.

airelil: We got back to what we started from :-)

SerovAlexey: @vasily-v-ryabov ,apparently, we have a different understanding of the timeout. 
tell me, please, how do I then check that timeut expired?
it is impossible to simulate this situation. in order to display the exception need at least a connection failure

SerovAlexey:
@vasily-v-ryabov @airelil ,check,please

pywinauto/application.py
```diff
@@ -845,9 +845,9 @@ def connect(self, **kwargs):
             try:
                 timeout = float(kwargs['timeout'])
                 del kwargs['timeout']
-                time.sleep(timeout)
+                wait_until_passes(timeout, Timings.app_connect_retry, partial(self.connect, **kwargs), (ValueError, ProcessNotFoundError))
```
vasily-v-ryabov: `RuntimeError` is also possible as you can see below in the `connect` method.

pywinauto/application.py
```diff
@@ -839,6 +840,15 @@ def __connect(self, **kwargs):
     def connect(self, **kwargs):
         """Connects to an already running process"""
         connected = False
+
+        if 'timeout' in kwargs:
```
vasily-v-ryabov: If no timeout in the method call parameters, we need to use `app_start_timeout` by default.

pywinauto/application.py
```diff
@@ -839,6 +840,15 @@ def __connect(self, **kwargs):
...
         """Connects to an already running process"""
         connected = False
+
+        if 'timeout' in kwargs:
+            try:
+                timeout = float(kwargs['timeout'])
+                del kwargs['timeout']
+                wait_until_passes(timeout, Timings.app_connect_retry, partial(self.connect, **kwargs), (ValueError, ProcessNotFoundError))
+            except Exception as e:
+                raise e
```
vasily-v-ryabov: It seems we don't need `try-except` block here because we do nothing with this error, just re-raising.

pywinauto/application.py
SerovAlexey: If we use the default value of this will lead to infinite loops

pywinauto/application.py
SerovAlexey: try-except for validation timeout

SerovAlexey:
@vasily-v-ryabov fix

SerovAlexey:
@vasily-v-ryabov added default timeout,check,please

vasily-v-ryabov:
The tests on AppVeyor are failing. Please take a look.

SerovAlexey:
You may suggest where to look

vasily-v-ryabov:
The test is failing on AppVeyor with
```
ERROR: Test that connect_() raises with invalid input
----------------------------------------------------------------------
  File "C:\projects\pywinauto\pywinauto\unittests\test_application.py", line 355, in test_connect_raises
    Application().connect)
  File "c:\python27\lib\unittest\case.py", line 473, in assertRaises
    callableObj(*args, **kwargs)
  File "C:\projects\pywinauto\pywinauto\application.py", line 876, in connect
    connected = wait_until_passes(timeout, Timings.app_connect_retry, partial(get_connected, self = self, **kwargs), (ValueError, ProcessNotFoundError, RuntimeError))
  File "C:\projects\pywinauto\pywinauto\timings.py", line 397, in wait_until_passes
    func_val = func(*args)
  File "C:\projects\pywinauto\pywinauto\application.py", line 869, in get_connected
    return connected
UnboundLocalError: local variable 'connected' referenced before assignment
```

SerovAlexey:
добрый вечер.

да, увидел, но это не мои тесты упали
можете подсказать, как можно посмотреть предыдущие builds?
потому что в моем пулл реквесте нет ни одного коммита у которого прошли бы
тесты.

2016-05-13 22:48 GMT+03:00 Vasily Ryabov <notifications@github.com>:

vasily-v-ryabov:
Алексей, ответ на письмо автоматически добавляет коммент на гитхаб. Давайте сразу объясню, а потом удалю все комменты, которые по-русски.

Вот пример [слегка нестабильных тестов](https://ci.appveyor.com/project/pywinauto/pywinauto/build/0.6.0_build587). Можно ткнуть на красные таски и увидеть внизу, что это, в основном, не найденные элементы. Такие ошибки чаще бывают, когда в таске стоит UIA_SUPPORT=TRUE.

В вашем случае, если вы найдёте в [истории билдов](https://ci.appveyor.com/project/pywinauto/pywinauto/history) свои коммиты, все подзадачи в билде - красные. Если пройти внутрь любой задачи, хотя бы некоторые упавшие тесты - точно ваши. Главное, докрутить до конца output'а и прочитать traceback'и.

Конкретно по последней ошибке вы банально не перенесли начальное значение `connected = False` во внутреннюю функцию. Это прямо по описанию очевидно. Нужно просто уметь читать. :)

vasily-v-ryabov:
Could you also please merge recent changes from the upstream/UIA to your fork? It will update the PR automatically. Here is [an instruction](https://help.github.com/articles/syncing-a-fork/).

PR #203, opened 2016-05-19
Add mouse position hook to hooks.py

ethanhs:
First off, great work by @MaxSamokhvalov on the hooks.py file. I know the example I gave was pretty bad (I wrote it as one of my first projects when I was just learning to program), this is so much better. I may even ~~steal~~ use it in my module if that is okay with you all (just as a stand alone module, for those who don't need all of pywinauto).

In this PR I just added the mouse position information, as I had that in my original code, and found it useful. It was easy enough to add so take it if you want the addition. This would allow for checks if the mouse is in a bounding box or if it clicks in said box, etc.

I also added a little to the example in the file.

Oh, and you might want to mention that one _cannot_ exit this with ctrl+c (before and after this PR). When I first wrote this code I was somewhat surprised that it wouldn't quit the loop, as I expect that to kill python when I run things from the console.

### DIFF
--- a/pywinauto/hooks.py
+++ b/pywinauto/hooks.py
@@ -297,7 +297,7 @@ if __name__ == "__main__":
 
         if isinstance(args, MouseEvent):
             if args.current_key == 'RButton' and args.event_type == 'key down':
-                print("Right button pressed at ({}, {})".format(args.mouse_x,args.mouse_y))
+                print("Right button pressed at ({0}, {1})".format(args.mouse_x,args.mouse_y))
 
             if args.current_key == 'WheelButton' and args.event_type == 'key down':
                 print("Wheel button pressed")
--- a/pywinauto/hooks.py
+++ b/pywinauto/hooks.py
@@ -28,9 +28,11 @@ class KeyboardEvent(object):
 
 class MouseEvent(object):
     """Is created when mouse event catch"""
-    def __init__(self, current_key=None, event_type=None):
+    def __init__(self, current_key=None, event_type=None, mouse_x=0, mouse_y=0):
         self.current_key = current_key
         self.event_type = event_type
+        self.mouse_x = mouse_x
+        self.mouse_y = mouse_y
 
 
 class Hook(object):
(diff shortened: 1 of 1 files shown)


airelil:
I think that the loop in ```listen``` method is still buggy because ```GetMessageW``` is a blocking call and there are no other messages coming to the listener except of ```WM_QUIT```. So basically, it's not looping at all. I think, ```PeekMessage``` should be used here for the real looping 
```python
 while self.mouse_is_hook or self.keyboard_is_hook:
            msg = windll.user32.GetMessageW(byref(message), 0, 0, 0)
```

vasily-v-ryabov:
This is just a first raw implementation of `hooks.py`. We still have a lot of plans: trying `PeekMessage` as @airelil mentioned, writing tests, maybe some API re-factoring, adding more global hooks like `WH_CBT` which might be useful in writing script recorder.

Of course, you are free to use this module under terms of BSD 3-clause license that we will use for pywinauto 0.6.0. I'll add the license text to the module caption this week since it has no dependencies on other pywinauto code for now.

ethanhs:
@airelil I tried changing the GetMessageW call to the PeekMessageW call, and it did not help. I am guessing there may be something else going on.

airelil:
@ethan, ```PeekMessage``` returns a boolean, it behaves differently from ```GetMessage```, also I don't think we need to leave the messages in the queue. The following worked for me:
```python
        PM_REMOVE = 1
        WM_QUIT = 0x0012
        while self.mouse_is_hook or self.keyboard_is_hook:

            res = windll.user32.PeekMessageW(byref(message), 0, 0, 0, PM_REMOVE)
            if res:
                if message.message == WM_QUIT:
                    hk.unhook_keyboard()
                    hk.unhook_mouse()
                    exit(0)

                else:
                    windll.user32.TranslateMessage(byref(message))
                    windll.user32.DispatchMessageW(byref(message))
```

vasily-v-ryabov:
See PR #204 for the license text.
Further improvements for `hooks.py` are very welcome.
@ethanhs, if you submit more PRs, we are happy to review them.

I'm currently working on `send_chars` method of `HwndWrapper` and have plans to switch on the documentation soon. @airelil is working on ListView support in UIA back-end. So `hooks.py` is not the highest priority for now, but generally it's an important feature.

PR #207, opened 2016-06-14
UIA optimizations

airelil:
1. Add to cache UIAElementInfo.rich_text as it's extensievely used during the control name resolution.
2. Do not pre-cache UIAElementInfo properties untill implicit call to the property
3. Small optimizations find_elements and get_non_text_control_name
4. Fix UIAWrapper._select to work with own methods instead of direct access to lower level element_info functions
5. Fix EOL in test_application.py

### DIFF
--- a/pywinauto/findwindows.py
+++ b/pywinauto/findwindows.py
@@ -193,6 +193,10 @@ def find_elements(class_name = None,
         if ctrl_index is not None:
             return [elements[ctrl_index], ]
 
+    # early stop
+    if not elements:
+        return elements
+
     if framework_id is not None and elements:
         elements = [elem for elem in elements if elem.framework_id == framework_id]
 
@@ -221,10 +225,6 @@ def find_elements(class_name = None,
                 break
         if not found_active:
             elements = []
-
-    # early stop
-    if not elements:
-        return elements
 
     if class_name is not None:
         elements = [elem for elem in elements if elem.class_name == class_name]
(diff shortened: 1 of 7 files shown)


pywinauto/findwindows.py
```diff
@@ -255,12 +257,14 @@ def _title_match(w):
         elements = [elem for elem in elements if elem.enabled]
 
     if best_match is not None:
+        # Build a list of wrapped controls.
+        # Speed up the loop by setting up local pointers
         wrapped_elems = []
+        add_to_wrp_elems = wrapped_elems.append
+        wrp_cls = backend_obj.generic_wrapper_class
```
vasily-v-ryabov: Is it really running faster? I thought module attribute resolution wouldn't be a bottleneck.

pywinauto/uia_element_info.py
```diff
@@ -95,15 +90,17 @@ def __init__(self, handle_or_elem = None):
...
+ 
+        self.set_cache_strategy(cache_enable)
 
     def _get_current_class_name(self):
         """Return an actual class name of the element"""
         return self._element.CurrentClassName
     
     def _get_cached_class_name(self):
         """Return a cached class name of the element"""
+        if self._cached_class_name == False:
```
vasily-v-ryabov: This comparison produces Landscape warning. Do you mean that `self._cached_class_name == "" or None` is not the case?

pywinauto/findwindows.py
airelil: It depends on calls. Mostly it doesn't change too much but sometimes you see the difference.
This is an example of line profiling in ipython on my py27 x64:
```
%lprun -f pywinauto.findwindows.find_elements dlg.TabControl.select("Views")
```

```
   262         2            7      3.5      0.0          wrapped_elems = []
   263        65          328      5.0      0.0          for elem in elements:
   264        63          212      3.4      0.0              try:
   265        63        31850    505.6      4.1                  wrapped_elems.append(backend_obj.generic_wrapper_class(

```
```

   264         2            8      4.0      0.0          wrp_cls = backend_obj.generic_wrapper_class
   265        65          329      5.1      0.0          for elem in elements:
   266        63          215      3.4      0.0              try:
   267        63        30654    486.6      2.9                  add_to_wrp_elems(wrp_cls(elem))
```

PR #208, opened 2016-06-19
UIA: fix switching off the temporary caching in find_elements

airelil:
Fix switching off the temporary caching in find_elements
Optimize lookup of a controls in test_UIAWrapper by specifying their class name


pywinauto/uia_element_info.py
```diff
@@ -161,7 +161,7 @@ def _get_cached_rich_text(self):
 
     def set_cache_strategy(self, cached = None):
         """Setup a cache strategy for frequently used attributes"""
-        if cached:
+        if cached == True:
```
vasily-v-ryabov: Could be `if cached is True` to avoid Landscape.io minor warning.

PR #211, opened 2016-07-10
Multi-level WindowSpecification objects (3+ levels in a chain)

vasily-v-ryabov:
* Allow 3+ levels in an access chain of attributes.
* Add method `texts()` into ListViewWrapper.
* Add class `Desktop` as an alternative root of WindowSpecification objects (without process info).
* Fix "copypasta" bug in `backend.py`.

### DIFF
--- a/pywinauto/controls/uia_controls.py
+++ b/pywinauto/controls/uia_controls.py
@@ -581,7 +581,7 @@ class ListViewWrapper(uiawrapper.UIAWrapper):
             hdr = None
 
         return hdr
-   
+
     #-----------------------------------------------------------
     def cell(self, row, column):
         """Return a cell in the ListView control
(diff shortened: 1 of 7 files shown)


pywinauto/controls/uia_controls.py
```diff
@@ -628,3 +628,10 @@ def get_item(self, row):
 
     item = get_item  # this is an alias to be consistent with other content elements
 
+    #-----------------------------------------------------------
+    def texts(self):
+        """Return a list of item texts"""
+        return [ch.window_text() for ch in self.children() if ch.element_info.element.CurrentIsContentElement]
```
airelil: Ideally the children should be filtered already filtered out with ```PropertyCondition``` ```UIA_IsContentElementPropertyId```

PR #212, opened 2016-07-15
UIA: more methods for ListViewWrapper

airelil:
More methods for ListViewWrapper: get_selected_count/get_item_rect/is_checked/columns/get_column
Remove UIA control types from win32 control wrappers

### DIFF
--- a/pywinauto/controls/uia_controls.py
+++ b/pywinauto/controls/uia_controls.py
@@ -546,12 +546,12 @@ class ListItemWrapper(uiawrapper.UIAWrapper):
 
     #-----------------------------------------------------------
     def is_checked(self):
-        """Return True if the ListItem is checked"""
-        # Only items supporting Toggle pattern should answer
-        try:
-            res = (self.iface_toggle.ToggleState_On == toggle_state_on)
-        except NoPatternInterfaceError:
-            res = False
+        """Return True if the ListItem is checked
+        
+        Only items supporting Toggle pattern should answer.
+        Raise NoPatternInterfaceError if the pattern is not supported
+        """
+        res = (self.iface_toggle.ToggleState_On == toggle_state_on)
         return res
 
 
--- a/pywinauto/controls/uia_controls.py
+++ b/pywinauto/controls/uia_controls.py
@@ -661,7 +661,8 @@ class ListViewWrapper(uiawrapper.UIAWrapper):
         """Return the bounding rectangle of the list view item
         
         The interface is kept mostly for a backward compatibility
-        with the native ListViewWrapper interface"""
+        with the native ListViewWrapper interface
+        """
         itm = self.get_item(item_index)
         return itm.rectangle()
 
--- a/pywinauto/controls/uia_controls.py
+++ b/pywinauto/controls/uia_controls.py
@@ -628,3 +628,16 @@ class ListViewWrapper(uiawrapper.UIAWrapper):
 
     item = get_item  # this is an alias to be consistent with other content elements
 
+    #-----------------------------------------------------------
+    def get_selected_count(self):
+        """Return the number of selected items"""
+        # The call can be quite expensieve as we retrieve all 
+        # the selected items in order to count them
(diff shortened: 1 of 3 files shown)


pywinauto/controls/uia_controls.py
```diff
@@ -543,6 +544,16 @@ def is_selected(self):
...
         return self.iface_selection_item.CurrentIsSelected
 
+    #-----------------------------------------------------------
+    def is_checked(self):
+        """Return True if the ListItem is checked"""
+        # Only items supporting Toggle pattern should answer
+        try:
+            res = (self.iface_toggle.ToggleState_On == toggle_state_on)
+        except NoPatternInterfaceError:
+            res = False
```
vasily-v-ryabov: Maybe we shouldn't return `False` silently in this case. I'd suggest to leave this error unhandled or re-raise kind of `RuntimeError`. It should be more clear for users.

pywinauto/unittests/test_UIAWrapper.py
```diff
@@ -657,7 +657,11 @@ def test_item_count(self):
 
         def test_column_count(self):
             """Test the columns count in the ListView control"""
-            self.assertEqual(self.ctrl.column_count(), len(self.texts[0]))
+            num_cols = self.ctrl.column_count()
+            self.assertEqual(num_cols, len(self.texts[0]))
+            col = self.ctrl.get_column(1)
+            col.texts()[0] == u'Name'
```
vasily-v-ryabov: Should it have `self.assertEqual`?

pywinauto/unittests/test_UIAWrapper.py
```diff
@@ -666,20 +670,27 @@ def test_get_header_control(self):
 
         def test_select(self):
             """Test selecting an item of the ListView control"""
+            # Verify get_selected_count
+            cnt = self.ctrl.get_selected_count()
+            self.assertEqual(cnt, 0)
```
vasily-v-ryabov: I think it should be fine as a one liner: `self.assertEqual(self.ctrl.get_selected_count(), 0)`

vasily-v-ryabov:
Recently I also observed some problems with `ItemsView.get_item("file_name").move_mouse()` for an `explorer.exe` file list. It should use `texts()` method to be more correct (header control may shift an index and make it incorrect). I had no time to fix this issue yet.

P.S. `ItemsView.get_item(1).move_mouse()` worked just fine.

airelil:
Regarding ```get_item```, it's based on lookup of descendants with the text argument as PropertyCondition for ```FindAll```, while ```texts``` method exploits ```DocumentRange``` pattern. I need to look into it further.

vasily-v-ryabov:
OK. I've implemented method `texts()` specially for ListView in previous PR if any.

vasily-v-ryabov:
The following unused code could be removed now from `win32_controls.py`:
```python
if sysinfo.UIA_support:
     from ..uia_defines import IUIA
```

pywinauto/controls/uia_controls.py
```diff
@@ -629,6 +657,26 @@ def get_item(self, row):
     item = get_item  # this is an alias to be consistent with other content elements
 
     #-----------------------------------------------------------
+    def get_item_rect(self, item_index):
+        """Return the bounding rectangle of the list view item
+        
+        The interface is kept mostly for a backward compatibility
+        with the native ListViewWrapper interface"""
```
vasily-v-ryabov: Closing triple double quotes should be on a separate line (landscape minor warning).

airelil:
BTW, it's also possible to use the name resolution to access a list item. Let's say we have explorer.exe with "Downloads" window title and we want to click on an item in the file list, named SkypeSetup.exe. The following works for me: ```app.Downloads.SkypeSetupExe.set_focus().click_input()```

vasily-v-ryabov:
Cool! I'll prepare `explorer.exe` example for future readme soon.

PR #219, opened 2016-08-14
Make six.py a dependency, add codecov.yml

vasily-v-ryabov:
* codecov.yml is now a requirement for codecov.io
* `six` is now imported as a dependency (TODO: setup.py still needs to be updated)


pywinauto/tests/truncation.py
```diff
@@ -148,7 +145,7 @@ def _GetMinimumRect(text, font, usableRect, drawFlags):
     """Return the minimum rectangle that the text will fit into
 
     Uses font, usableRect and drawFlags information to find how
-    how to do it accurately
+    how to do it accurately.
```
airelil: A small typo: " find how how to..."

PR #221, opened 2016-08-31
Update all the license headers and make them unified

vasily-v-ryabov:
There is no need to add individual copyrights any more into some flexible number of files, all the contributions are now referenced as it's done by GitHub.

TODO: will update the credits page later.


airelil:
There some ```<br>``` html formatting tags in pywinauto/docs/credits.txt that have to be removed

image

vasily-v-ryabov:
@airelil, sure. Will do that separately.

PR #225, opened 2016-09-06
Update credits page, fix major landscape.io warnings.

vasily-v-ryabov:


airelil:
All good. Just small grammar typos like: "```The``` main layout..", "```A``` special layout..." ".. ```a``` combination of 64-bit Python and 32-bit app...."

Issue #100, opened 2015-10-08
RowList.exe toolbar texts and tool tips' texts are not available


vasily-v-ryabov:
```python
>>> t = app.RowListSampleApplication.Toolbar.WrapperObject()
>>> t.ButtonCount()
10L
>>> t.Texts()
['', u'', u'', u'', u'', u'', u'', u'', u'', u'', u'']
>>> t.TipTexts()
Traceback (most recent call last):
  File "<interactive input>", line 1, in <module>
  File "C:\...\pywinauto\pywinauto\controls\common_controls.py", line 2289, in TipTexts
    if not (-1 <= btn_tooltip_index < self.GetToolTipsControl().ToolCount()):
  File "C:\...\pywinauto\pywinauto\controls\common_controls.py", line 2323, in GetToolTipsControl
    return ToolTipsWrapper(self.SendMessage(win32defines.TB_GETTOOLTIPS))
  File "C:\...\pywinauto\pywinauto\controls\HwndWrapper.py", line 177, in __new__
    obj.__init__(handle)
  File "C:\...\pywinauto\pywinauto\controls\common_controls.py", line 2673, in __init__
    super(ToolTipsWrapper, self).__init__(hwnd)
  File "C:\...\pywinauto\pywinauto\controls\HwndWrapper.py", line 205, in __init__
    raise InvalidWindowHandle(hwnd)
InvalidWindowHandle: Handle 0x0 is not a vaild window handle
>>> t.Class()
u'ToolbarWindow32'
```
It would be great to find a root cause of the problem. It looks like standard MFC toolbar. If it has some unusual style or it's another corner case, we need to handle it as well.

vasily-v-ryabov:
It looks like pywinauto support only `CToolBarCtrl` class in MFC (CmnCtrl1.exe).
`CToolBar`/`CMFCToolBar` (RowList.exe/RebarTest.exe) requires more efforts to support.

vasily-v-ryabov:
There is useful article about MFC Feature Pack (see about `MFC Feature Pack samples` at the end):
http://www.codeproject.com/Articles/217588/Porting-a-legacy-MFC-app-to-MFC-Feature-Pack

vasily-v-ryabov:
It seems [TB_GETBUTTONTEXT](https://msdn.microsoft.com/en-us/library/windows/desktop/bb787325(v=vs.85).aspx) is not in use now. But it's a recommended way to retrieve a button text.

[...]

PR #232, opened 2016-09-25
UIA: ToolbarWrapper, cleanups

airelil:
- ToolbarWrapper for UIA backend (solves issue #100). Only methods complement to the methods of the base class are implemented.
- common_controls cleanup, remove trailing spaces

### DIFF
--- a/pywinauto/controls/uia_controls.py
+++ b/pywinauto/controls/uia_controls.py
@@ -848,3 +848,30 @@ class MenuWrapper(uiawrapper.UIAWrapper):
             raise IndexError()
 
         return menu
+
+
+# ====================================================================
+class ToolbarWrapper(uiawrapper.UIAWrapper):
+
+    """Wrap an UIA-compatible ToolBar control"""
+
+    _control_types = [
+        IUIA().UIA_dll.UIA_ToolBarControlTypeId
+    ]
+
+    # -----------------------------------------------------------
+    def __init__(self, elem):
+        """Initialize the control"""
+        super(ToolbarWrapper, self).__init__(elem)
+
+    @property
+    def writable_props(self):
+        """Extend default properties list."""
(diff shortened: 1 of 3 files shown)


pywinauto/controls/uia_controls.py
```diff
@@ -848,3 +848,108 @@ def item_by_path(self, path, exact=False):
...
+    def button(self, button_identifier, exact=True, by_tooltip=False):
+        """Return the button at index button_index"""
+
+        cc = self.children()
+        texts = [c.window_text() for c in cc]
+        if isinstance(button_identifier, six.string_types):
+            self.actions.log('Toolbar buttons: ' + str(texts))
+
+            if by_tooltip:
+                texts = self.tip_texts()
```
vasily-v-ryabov: I can't find method `tip_texts()` in `UIAWrapper` or its derivatives.

pywinauto/unittests/test_UIAWrapper.py
```diff
@@ -25,19 +25,28 @@
...
+   os.path.dirname(__file__), r"..\..\apps\MFC_samples")
+if is_x64_Python():
+    mfc_samples_folder = os.path.join(mfc_samples_folder, 'x64')
+
 if UIA_support:
 
+    def _set_timings():
+        """Setup timings for UIA related tests"""
+        Timings.Defaults()
+        Timings.window_find_timeout = 20
```
vasily-v-ryabov: Agree. It's better to have this number in one place.

pywinauto/controls/uia_controls.py
airelil: I missed that one with "copy-paste". However, I think it has to be removed as I can't see a clean method in UIA for retrieving tooltips. So far only hovering mouse over the button triggers drawing the tooltip window.

PR #242, opened 2016-10-03
Refactored ListView wrapper

cetygamer:
Since WPF ListBox, DataGrid and ListView are basically the same thing, new `ListViewWrapper` supports them all.
It is difficult to make separate wrappers for DataGrid Control Type and List Control Type because on the one hand in WPF applications List may support Grid pattern and on the other hand DataGrid may represent a list control.
@vasily-v-ryabov, @airelil, please read code carefully, since there is a lot going on.


airelil:
As for me, ```is_content_element``` is a bit longer name than I would use, also some lines are too long, but I can live with it.

pywinauto/base_wrapper.py
```diff
@@ -336,31 +336,33 @@ def texts(self):
         return texts_list
 
     #-----------------------------------------------------------
-    def children(self, class_name = None, title = None, control_type = None):
+    def children(self, class_name=None, title=None, control_type=None, is_content_element=None):
```
vasily-v-ryabov: I would name it `content_only` to be consistent with `top_level_only` in `find_windows`.

pywinauto/unittests/test_application.py
```diff
@@ -994,7 +994,7 @@ def tearDown(self):
         def test_folder_list(self):
             """Test that ListViewWrapper returns correct files list in explorer.exe"""
             files_list = self.desktop.MFC_samplesDialog.Shell_Folder_View.Items_View.wrapper_object()
-            self.assertEqual(files_list.texts(),
+            self.assertEqual([item.window_text() for item in files_list.get_items()],
```
vasily-v-ryabov: `files_list.texts()` does almost the same. Though it doesn't make difference because it's test code.

PR #250, opened 2016-10-21
Docs fixes

airelil:
Add links to find_elements function description to various methods using the same parameters.
Update HowTo.txt to reflect the latest methods names, rephrasing and small typos
Use the new style methods names in getting_started.txt
Additional PEP-8 fixes in findwindows module

### DIFF
--- a/docs/HowTo.txt
+++ b/docs/HowTo.txt
@@ -384,7 +384,7 @@ One way to get this is to do the following
   ::
 
    import pywinauto.application
-   app = pywinauto.application.Application().connect_(path="explorer")
+   app = pywinauto.application.Application().connect(path="explorer")
    systray_icons = app.ShellTrayWnd.NotificationAreaToolbar
 
 The taskbar module provides very preliminary access to the System Tray.
--- a/docs/HowTo.txt
+++ b/docs/HowTo.txt
@@ -2,48 +2,48 @@
 How To's
 ========
 
-How to specify an usable Application instance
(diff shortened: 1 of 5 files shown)


docs/HowTo.txt
```diff
@@ -345,54 +345,54 @@ would use
...
+Toolbar control with a title "Notification Area".
 
 One way to get this is to do the following 
 
   ::
 
-   imprt pywinauto.application
-   app = pywinauto.application.Application().connect_(path = "explorer")
+   import pywinauto.application
+   app = pywinauto.application.Application().connect_(path="explorer")
```
vasily-v-ryabov: Maybe using `connect` is better than `connect_`.

pywinauto/application.py
```diff
@@ -847,7 +847,19 @@ def __connect(self, **kwargs):
...
+        The action is performed according to only one of parameters
+
+        :param process: a process ID of the target
+        :param handle: a window handle of the target
+        :param path: a path used to launch the target
+
+        .. seealso::
+
+           :func:`pywinauto.findwindows.find_elements` - the keyword arguments that
+           are also can be used instead of **process**, **handle** or **path**
```
vasily-v-ryabov: I would say `instead of or in addition to`.
Example: `app.connect(path='explorer.exe', title='Program Files')`

airelil: I wish it worked that way, but it doesn't.

PR #251, opened 2016-10-23
Make deprecated API available for backend="win32" only

vasily-v-ryabov:
* Enable Codacy coverage submission (the results are not aggregated by multiple runs yet).
 * Improve `print_control_identifiers()`.
 * Make `children()` and `descendants()` process-agnostic by default (useful for Win10 Calculator example). Yes, some windows in Store apps can have children in different processes! So we could miss them even if using `Desktop(backend"uia")` object as a root.
 * Enable UIA tests for 32-bit Python 3.5.
 * Enable FuzzyDict tests in a separate file.

Surprisingly FuzzyDict tests are a bit unstable (2 failures in ~2 out of 8 runs).
`testCheckBoxes` for native TreeView is also painful area. :(


pywinauto/application.py
```diff
@@ -556,19 +550,21 @@ def _ctrl_identifiers(self):
 
         return control_name_map
 
-    def print_control_identifiers(self, depth = 2):
+    def print_control_identifiers(self, depth = None):
```
airelil: Printing the whole hierarchy by default looks a bit exaggerated for me. User might get confused with a huge number of prints spat out by the function. Maybe we could leave ```depth=2``` but print an info before the all identifiers that there are other options if more levels are required. Something like:
```"Print out identifiers for {0} levels. Use 'depth' argument to get info about deeper sub-levels".format(depth)```

File: pywinauto/unittests/test_fuzzydict.py [hunk 2]
```diff
@@ -0,0 +1,99 @@
...
+
+class FuzzyTestCase(unittest.TestCase):
+
+    """Perform some tests"""
+
+    test_dict = {
+        'Hiya'  : 1,
+        'hiy\xe4' : 2,
+        'test3' : 3,
+        1: 324}
```
airelil: I think if the dictionary strings will be declared explicitly as unicode strings the search order of "test_dict" won't change between different python versions
```
        test_dict = {
            u'Hiya'  : 1,
            u'hiy\xe4' : 2,
            u'test3' : 3,
            1: 324}
```

vasily-v-ryabov: Will try, though `from __future__ import unicode_literals` should do the trick but it didn't.

PR #272, opened 2016-11-21
Fix issue 270: Replace AttachThreadInput with ShowWindow in set_focus()

mtkennerly:
```AttachThreadInput``` fails when the foreground window does not have a message loop, namely cmd.exe. Using ```ShowWindow``` instead avoids the need for a lot of the existing handling to juggle threads.

While it was suggested in #270 that the ```mouse.move()``` calls may cause problems as well, I left it alone since it was outside the scope of this specific issue.

### DIFF
--- a/pywinauto/controls/hwndwrapper.py
+++ b/pywinauto/controls/hwndwrapper.py
@@ -1169,21 +1169,26 @@ class HwndWrapper(BaseWrapper):
         (https://msdn.microsoft.com/en-us/library/windows/desktop/ms633539(v=vs.85).aspx)
         so the mouse cursor is removed from the screen to prevent any side effects.
         """
-        # Notice that we need to move the mouse out of the screen
-        # but we don't use the built-in methods of the class:
-        # self.mouse_move doesn't do the job well even with absolute=True
-        # self.move_mouse_input can't be used as it calls click_input->set_focus
-        mouse.move(coords=(10000, 20000))
+        # "steal the focus" if there is another active window
+        # otherwise it is already into the foreground and no action required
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: pywinauto/controls/hwndwrapper.py [hunk 1]
```diff
@@ -1173,68 +1173,17 @@ def set_focus(self):
...
 
-            # make sure that we are idle before returning
-            win32functions.WaitGuiThreadIdle(self)
+        # make sure that we are idle before returning
+        win32functions.WaitGuiThreadIdle(self)
 
-            # only sleep if we had to change something!
-            time.sleep(Timings.after_setfocus_wait)
+        # only sleep if we had to change something!
+        time.sleep(Timings.after_setfocus_wait)
```
airelil: @mtkennerly, thanks for the PR. Even if we decide to stop stealing focus with ```AttachThreadInput``` it's still required to verify if the window already has a focus. I think that there is no point in running ``` ShowWindow/SetForeground/time.sleep``` sequence bluntly. I would also still set mouse move coordinates different from the left bottom corner.
```python
cur_foreground = win32gui.GetForegroundWindow()
if self.handle != cur_foreground:
    mouse.move(coords=(-10000, 500))  # move the mouse out of screen to the right
    win32gui.ShowWindow(self.handle, win32con.SW_RESTORE)
    win32gui.SetForegroundWindow(self.handle)
    win32functions.WaitGuiThreadIdle(self)
    time.sleep(Timings.after_setfocus_wait)
```

vasily-v-ryabov:
Good job! The tests look stable.
Merging.

PR #273, opened 2016-11-23
win32_hooks refactoring and unit tests

airelil:
win32_hooks: 
- refactor events loop, low-level handlers
- fix warnings on return from "finally"
- more constants in ID_TO_KEY dictionary
- add a test suite for win32_hooks
  - test with real keys/mouse through SendKeys/mouse_click
  - test with mocking system dependencies

win32defines: 
- group and sort VK_ and WM_ constants


pywinauto/win32_hooks.py
```diff
@@ -241,83 +407,143 @@ class Hook(object):
...
-                    event_type = self.event_types[0xFFFFFFFF & event_code]
-
-                    if event_type == 'key down':
-                        self.pressed_keys.append(current_key)
-
-                    if event_type == 'key up':
-                        self.pressed_keys.remove(current_key)
-
-                    event = KeyboardEvent(current_key, event_type, self.pressed_keys)
+            @HOOKCB
```
vasily-v-ryabov: Wow! Didn't know `CFUNCTYPE` object can be used as a decorator. 👍

pywinauto/win32defines.py
```diff
@@ -34,23 +34,21 @@
 
 """Definition of Windows defines"""
 from __future__ import unicode_literals
+from ctypes import c_int
 
 __revision__ = "$Revision$"
```
vasily-v-ryabov: I think these `__revision__` lines could be removed. It's inherited from Mercurial or SVN probably.

pywinauto/win32_hooks.py
```diff
@@ -241,83 +407,143 @@ class Hook(object):
...
+                al = ActionLogger()
+                al.log("_process_kbd_msg_type, can't remove a key: {0}".format(current_key))
+
+        return event_type
+
+    def _keyboard_ll_hdl(self, code, event_code, kb_data_ptr):
+        """Execute when a keyboard low level event was catched"""
+        try:
+            # The next hook in chain must be always called
+            res = windll.user32.CallNextHookEx(self.keyboard_id,
```
airelil: The MSDN docs say: "A hook procedure can call this function either before or after processing the hook information." Notice that the hooks run in the context of an application so, I think, the ```CallNextHookEx``` simply sends a message to another registered app (we by ourselves have the loop message) Maybe it also calls to ```LoadLibrary```.

airelil:
Regarding pywin32 dependency. It's an interesting point. We don't see any users complaining about a burden of this dependency and we already have additional one - comtypes. I'd remove it only if the package is stopped be actively maintained or if it prevents from migrating to Python3

vasily-v-ryabov:
Great! Now everything looks fine. Thanks a lot!

PR #274, opened 2016-11-24
Update HowTo.txt

eromoe:
fix broken code from
`app = Application.connect(path=r"c:\windows\system32\notepad.exe")`
to
`app = Application().connect(path=r"c:\windows\system32\notepad.exe")`


vasily-v-ryabov:
Thanks for the fix!

Issue #269, opened 2016-11-18
 TypeError: adding NoneType + String in findbestmatch.py


halfnibble:
I'm not sure if it's just my code, but this seems like something that should be catched.

I'm getting the following error on 0.6.0 with certain child_window selections:
```
pywinauto\findbestmatch.py", line 274, in get_non_text_control_name
    best_name = text_ctrl.window_text() + ctrl_friendly_class_name
TypeError: unsupported operand type(s) for +: 'NoneType' and 'str'
```

I'm wondering if there should be a try/catch block around the statement at [line 274](https://github.com/pywinauto/pywinauto/blob/master/pywinauto/findbestmatch.py#L274)?

halfnibble:
Interestingly, if I wrap my problem selector with in an if statement:
```
if app.<selector>.exists():
    app.<selector>.click()
```
Then the TypeError goes away and the click action is performed.

airelil:
Noninitialized friendly_class_name definitely looks as a bug. What type of backend do you use?

airelil:
Oh. Probably it's not friendly_class_name, but text_ctrl.window_text()

halfnibble:
I'm using 'win32'. For a legacy VB app that was probably made before I was born...

I think it's text_ctrl.window_text(). But I just started looking into the codebase a moment ago.

airelil:
Maybe you'll find this oneliner be useful for you :
```app.selector.wait('exists').click()```

You can add an optional timeout parameter as well.

halfnibble:
@airelil Actually, using `wait()` also causes the TypeError: NoneType + String bug.

It only works with `if app.<selector>.exists():`. I wish I could provide better debugging info. All I can think is that the control tree is so jacked up that searching for a best match is somehow coming across an unanticipated error state.

halfnibble:
Would providing a .txt file with the list of controls help in any way?

halfnibble:
I then ran across an issue whereby my "if exists()" fix wasn't working either. This may have something to do with the fact I'm running stuff in a **for** loop. I don't know. I hacked it into working by setting `time.sleep(1)` before my **for** loop starts.

halfnibble:
So, I'm going to close this issue for most likely not affecting other users.

vasily-v-ryabov:
Hi @halfnibble, the code in `findbestmatch` is not safe. I'll amend. Thanks for reporting this!

PR #279, opened 2016-12-21
Small fixes and optimizations

### LINKED ISSUE #256: Any possibility to make backend uia faster
i have a very simple script that click few buttons in calculator app. If I use backend win32 executing script takes about 16 sec, but if I use backend uia same script takes 1 min 35 sec. Below examples scripts. Each script is executed 4 times.

script below uses buttons text to find it

```
app = Application()
appconnect = app.connect(title='Calculator')
app_dlg = appconnect.top_window_()
app_dlg.window_(best_match='3').Click()
```

script below uses buttons automatinID

```
app = Application(backend='uia')
appconnect = app.connect(title='Calculator')
app_dlg = appconnect.top_window_()
app_dlg.child_window(auto_id='button5').click_input()
```

vasily-v-ryabov:
* Fix crash #269 
* Partially fix #256 with small optimizations
* Cosmetic changes

Had no chance to write more tests yet.


pywinauto/base_wrapper.py
```diff
@@ -797,11 +797,13 @@ def type_keys(
         .. _keyboard: pywinauto.keyboard.html
         """
         self.verify_actionable()
+        friendly_class_name = self.friendly_class_name()
 
         if pause is None:
             pause = Timings.after_sendkeys_key_wait
 
         if set_foreground:
+            # self.top_level_parent().set_focus()
```
airelil: Debug line ? Should be removed

pywinauto/findwindows.py
```diff
@@ -200,10 +201,11 @@ def find_elements(class_name=None,
         # if not given a parent look for all children of the desktop
         if not parent:
             parent = backend_obj.element_info_class()
+        elif isinstance(parent, BaseWrapper):
+            parent = parent.element_info
```
airelil: I'm not sure I understand it. Is the check for the instance of ```backend_obj.generic_wrapper_class``` at line 180 not enough ?

pywinauto/unittests/test_application.py
```diff
@@ -434,12 +434,12 @@ def _cabinetwclass_exist():
 
     def test_wait_cpu_usage_lower_uia(self):
         """Test that wait_cpu_usage_lower() works correctly for UIA"""
-        app = Application(backend='uia')
-        app.start('notepad.exe')
+        self.app = Application(backend='uia')
+        self.app.start('notepad.exe')
```
airelil: A minor note. Why ```app``` as a local variable was not enough ?

airelil:
Oh. I just noticed a number of faliures on Appveyor. ```test_wait_cpu_usage_lower_uia``` should be run conditionally only when UIA backend is enabled

pywinauto/unittests/test_application.py
vasily-v-ryabov: It adds a Codacy warning. But it's minor. Will try to revert back and see if this check is red or not.

airelil:
Maybe we need to use mock more extensievly If we want to increase the coverage. For example:
```
        while self.cpu_usage(usage_interval) > threshold:
            if time.time() - start_time > timeout:
                raise RuntimeError('Waiting CPU load <= {}% timed out!'.format(threshold))
```
can be easily tested with mocking ```self.cpu_usage``` method

PR #299, opened 2017-02-03
Send WM_CHAR when possible during a send_keys

batterseapower:
In particular this is necessary for the enter key to work with some apps (they expect a WM_CHAR with keycode 13, not just the WM_KEYDOWN/UP).

Enter still doesn't work in all situations (e.g. the commented out test_send_chars_enter still fails) but this is an improvement on the status quo, I think.

### DIFF
--- a/pywinauto/controls/hwndwrapper.py
+++ b/pywinauto/controls/hwndwrapper.py
@@ -470,40 +470,39 @@ class HwndWrapper(BaseWrapper):
         For more information about Sequences and Modifiers navigate to keyboard.py
         """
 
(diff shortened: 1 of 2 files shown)


vasily-v-ryabov:
Great! Thank you for improving this method.

There is only one minor warning produced by Codacy: https://www.codacy.com/app/pywinauto/pywinauto/pullRequest?prid=503108
`hwndwrapper.py, line 481`

But I'd prefer to prevent growing number of warnings, so any new warning should be fixed before PR merge. Thanks!

batterseapower:
Thanks for the info! Looks like my latest commit solved the warning.

PR #301, opened 2017-02-06
Example for Win10 Calculator and setup.py fix

vasily-v-ryabov:
Hope this PR will close 0.6.1 milestone.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,5 +1,5 @@
 # GUI Application automation and testing library
-# Copyright (C) 2006-2016 Mark Mc Mahon and Contributors
+# Copyright (C) 2006-2017 Mark Mc Mahon and Contributors
 # https://github.com/pywinauto/pywinauto/graphs/contributors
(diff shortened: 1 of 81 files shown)


setup.py
```diff
@@ -57,14 +57,19 @@ def setup_path(path = ""):
 #        shutil.move(setup_path("website"), setup_path("docs"))
 
 if sys.platform == 'win32':
-    install_requires = ['six', 'pypiwin32', 'comtypes']
+    install_requires = ['six', 'comtypes']
+    try:
+        import win32api # check if it was already installed manually
+    except ImportError:
+        install_requires.append('pypiwin32')
+
```
airelil: I'm a bit puzzled here. Isn't that should be handled by pip automatically ? If not, what about the other dependencies: comtypes and six ?

vasily-v-ryabov: If you install pyWin32 manually from SourceForge, it's called `pywin32`, but on PyPI it's called `pypiwin32`. That's what makes pip (and people) crazy. :)

PR #339, opened 2017-04-07
master

maollm:
fix call wait_not/wait error: not method Exists found (UIA)


vasily-v-ryabov:
Thanks for finding and fixing this! 👍

PR #343, opened 2017-04-17
Add output to file in print_control_identifiers()

sovrasov:
- Add `filename` parameter
- Add unit tests
- Add aliases for `print_control_identifiers()`: `print_ctrl_ids` and `dump_tree`

### DIFF
--- a/pywinauto/application.py
+++ b/pywinauto/application.py
@@ -583,14 +583,6 @@ class WindowSpecification(object):
         for name, control in name_control_map.items():
             control_name_map.setdefault(control, []).append(name)
 
-        log_func = print
-        log_file = None
-        if filename is not None:
-            log_file = open(filename, "w")
-            log_func = lambda arg: log_file.write(str(arg) + os.linesep)
-
-        log_func("Control Identifiers:")
-
         def print_identifiers(ctrls, current_depth=1, log_func=print):
             """Recursively print ids for ctrls and their descendants in a tree-like format"""
             if len(ctrls) == 0 or current_depth > depth:
@@ -640,9 +632,15 @@ class WindowSpecification(object):
 
                 print_identifiers(ctrl.children(), current_depth + 1, log_func)
 
-        print_identifiers([this_ctrl, ], log_func=log_func)
-
-        if log_file is not None:
+        if filename is None:
+            print("Control Identifiers:")
+            print_identifiers([this_ctrl, ])
+        else:
+            log_file = open(filename, "w")
+            def log_func(msg):
+                log_file.write(str(msg) + os.linesep)
+            log_func("Control Identifiers:")
+            print_identifiers([this_ctrl, ], log_func=log_func)
             log_file.close()
 
     print_ctrl_ids = print_control_identifiers
--- a/pywinauto/application.py
+++ b/pywinauto/application.py
@@ -583,14 +583,15 @@ class WindowSpecification(object):
         for name, control in name_control_map.items():
             control_name_map.setdefault(control, []).append(name)
 
+        log_func = print
         log_file = None
-        if filename is None:
-            print("Control Identifiers:")
-        else:
+        if filename is not None:
             log_file = open(filename, "w")
-            log_file.write("Control Identifiers:")
+            log_func = log_file.write
 
-        def print_identifiers(ctrls, current_depth=1, log_file=None):
+        log_func("Control Identifiers:")
+
+        def print_identifiers(ctrls, current_depth=1, log_func=print):
             """Recursively print ids for ctrls and their descendants in a tree-like format"""
             if len(ctrls) == 0 or current_depth > depth:
                 return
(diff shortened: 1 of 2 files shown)


pywinauto/application.py
```diff
@@ -626,15 +631,27 @@ def print_identifiers(ctrls, current_depth=1):
                     criteria_texts.append(u'control_type="{}"'.format(control_type))
                 if title or class_name or auto_id:
                     output += indent + u'child_window(' + u', '.join(criteria_texts) + u')'
-                if six.PY3:
-                    print(output)
+
+                if log_file is None:
```
vasily-v-ryabov: Looks like too many `if`s for me. How about one `if` and re-using something like `log_func` callable object (`log_func=print` by default).

pywinauto/unittests/test_application.py
```diff
@@ -886,6 +886,29 @@ def test_print_control_identifiers(self):
...
         self.ctrlspec.print_control_identifiers()
 
+    def test_print_control_identifiers_file_output(self):
+        """Make sure print_control_identifiers() creates correct file"""
+        output_filename = "test_print_control_identifiers.txt"
+        self.dlgspec.print_ctrl_ids(filename=output_filename)
+        if os.path.isfile(output_filename):
+            with open(output_filename, "r") as test_log_file:
+                content = str(test_log_file.readlines())
+                self.assertTrue(content.find("['Untitled - NotepadEdit', 'Edit']") != -1)
```
vasily-v-ryabov: The test fails here on AppVeyor:
```python
  File "C:\projects\pywinauto\pywinauto\unittests\test_application.py", line 896, in test_print_control_identifiers_file_output
    self.assertTrue(content.find("['Untitled - NotepadEdit', 'Edit']") != -1)
AssertionError: False is not true
```

BTW using `find` function is an old ugly style from 90s. Use operator `in`.

pywinauto/application.py
```diff
@@ -583,14 +583,15 @@ def print_control_identifiers(self, depth=None, filename=None):
...
 
+        log_func = print
         log_file = None
-        if filename is None:
-            print("Control Identifiers:")
-        else:
+        if filename is not None:
             log_file = open(filename, "w")
-            log_file.write("Control Identifiers:")
+            log_func = log_file.write
```
vasily-v-ryabov: `.write` method will not add line endings (`\n`) as print will do.

sovrasov: Can I use lambda here?
Something like `log_func = lambda arg: log_file.write(str(arg) + "\n")`

vasily-v-ryabov: Sorry, forgot to say. It will produce static warning since it's a named function so using `def log_func` is preferable in this condition branch.

pywinauto/unittests/test_application.py
```diff
@@ -886,6 +886,29 @@ def test_print_control_identifiers(self):
...
 
+    def test_print_control_identifiers_file_output(self):
+        """Make sure print_control_identifiers() creates correct file"""
+        output_filename = "test_print_control_identifiers.txt"
+        self.dlgspec.print_ctrl_ids(filename=output_filename)
+        if os.path.isfile(output_filename):
+            with open(output_filename, "r") as test_log_file:
+                content = str(test_log_file.readlines())
+                self.assertTrue("['Untitled - NotepadEdit', 'Edit']" in content
+                    or "['Edit', 'Untitled - NotepadEdit']" in content)
```
sovrasov: In python 2.7 u letter is added before control identifiers: `u'Untitled - NotepadNotepad'`

pywinauto/application.py
```diff
@@ -587,7 +587,7 @@ def print_control_identifiers(self, depth=None, filename=None):
         log_file = None
         if filename is not None:
             log_file = open(filename, "w")
-            log_func = log_file.write
+            log_func = lambda arg: log_file.write(str(arg) + os.linesep)
```
vasily-v-ryabov: Should be
```python
def log_func(msg):
    log_file.write(str(msg) + os.linesep)
```
According to this warning: https://www.quantifiedcode.com/app/project/gh:pywinauto:pywinauto/diff/fe70a29febedfc009f9c9ad88077fac15a3372ba/a7813da1ff743032c52dfd38ee06029a641cf555/issues

sovrasov: I've already tested variant with `def`. It causes a critical warning related to redefinition of `log_func` identifier (it had been already defined few lines upper). I'll move `if` which resolves log_finc after `def print_identifiers` to avoid this problem.

pywinauto/unittests/test_application.py
```diff
@@ -886,6 +886,29 @@ def test_print_control_identifiers(self):
...
+        self.dlgspec.print_ctrl_ids(filename=output_filename)
+        if os.path.isfile(output_filename):
+            with open(output_filename, "r") as test_log_file:
+                content = str(test_log_file.readlines())
+                self.assertTrue("'Untitled - NotepadEdit'" in content
+                    and "'Edit'" in content)
+                self.assertTrue("child_window(class_name=\"msctls_statusbar32\")" in content)
+            os.remove(output_filename)
+        else:
+            self.self.fail("print_control_identifiers can't create a file")
```
vasily-v-ryabov: `self.self.fail` :)
And below too. Though never executed.

PR #344, opened 2017-04-18
add feature

dd-han:


dd-han:
create PR for review, still need cleanup and better way for this

vasily-v-ryabov:
Probably you need using method `.descendants()` that returns the whole subtree as a plain list.
Or do you need specifically a tree-like nested list?

dd-han:
I did not find in the document, thank you for the tips.

Because the pressure on time, this is the few times to confirm whether there is a similar method. Would like to ask the current addition to the sample outside the more complete documents can refer to it? Or directly from the code to find the required method faster than looking for documents?

vasily-v-ryabov:
The docs are not ideal. I wrote the Getting Started Guide for 0.6.0+ which improved the situation but it's not complete as well. Hope to find time for the docs this summer. So more feedback is better. Will analyze issues tagged as "questions" before writing new docs.

PR #345, opened 2017-04-18
add logging to mspaint example

ArtemSkrebkov:


ArtemSkrebkov:
@vasily-v-ryabov 
I'm sorry but I can't understand what is wrong with code coverage.
Could you explain what is happen and  how to fix it?

examples/mspaint.py
```diff
@@ -8,15 +8,36 @@
 The example shows how to work with MS Paint application. It opens
 JPEG image and resizes it using "Resize and Skew" dialog.
 """
-
+import logging
```
vasily-v-ryabov: If you use `print` function please do `from __future__ import print_function` as a first import to be Py2/Py3 compatible.

examples/mspaint.py
```diff
@@ -8,15 +8,36 @@
...
+if len(sys.argv) >= 2 and len(sys.argv) <= 3:
+    if sys.argv[1] == '--log':
+        is_logging_enabled = True
+        file_name_log = 'log_mspaint.txt'
+        if len(sys.argv) == 3:
+            file_name_log = sys.argv[2]
+    else:
+        print("""Usage: python mspaint.py --log <filename>\n
+             --log - enable logging to a file (an optional parameter)\n
+             <filename> - log file name (an optional parameter, default: """)
```
vasily-v-ryabov: And so... What is the default behavior? Looks like missed line.

examples/mspaint.py
```diff
@@ -8,15 +8,36 @@
...
-
+import logging
+import sys
+from pywinauto import actionlogger
 from pywinauto import Application
 
+is_logging_enabled = False
+file_name_log = ''
+if len(sys.argv) >= 2 and len(sys.argv) <= 3:
+    if sys.argv[1] == '--log':
```
vasily-v-ryabov: Using `argparse.ArgumentParser` would look better.

examples/mspaint.py
```diff
@@ -8,15 +8,36 @@
...
+        is_logging_enabled = True
+        file_name_log = 'log_mspaint.txt'
+        if len(sys.argv) == 3:
+            file_name_log = sys.argv[2]
+    else:
+        print("""Usage: python mspaint.py --log <filename>\n
+             --log - enable logging to a file (an optional parameter)\n
+             <filename> - log file name (an optional parameter, default: """)
+        exit()
```
vasily-v-ryabov: Default behavior should just print all the data to stdout instead of specified file. Why exit?

examples/mspaint.py
```diff
@@ -8,15 +8,36 @@
...
+
+if is_logging_enabled:
+    actionlogger.enable()
+    logger = logging.getLogger('pywinauto')
+    logger.handlers[0].stream = open(file_name_log, 'w')
```
vasily-v-ryabov: Why not using standard class `FileHandler` explicitly? It should be more clear and you don't have to close anything at the end (the class will do this job for you).

examples/mspaint.py
```diff
@@ -9,14 +9,30 @@
 JPEG image and resizes it using "Resize and Skew" dialog.
 """
 
+from __future__ import print_function
+import logging
+import sys
```
vasily-v-ryabov: You have static analyzer warning here: sys in unused. Please remove this import.

examples/mspaint.py
```diff
@@ -9,14 +9,30 @@
...
+from pywinauto import actionlogger
 from pywinauto import Application
 
+import argparse
+parser = argparse.ArgumentParser()
+parser.add_argument("--log", help = "enable logging", type=str, required = False)
+args = parser.parse_args()
+
+file_name_log = args.log
```
vasily-v-ryabov: Why not use `args.log` directly below?

pywinauto/controls/uiawrapper.py
```diff
@@ -420,6 +420,11 @@ def invoke(self):
         """An interface to the Invoke method of the Invoke control pattern"""
         self.iface_invoke.Invoke()
 
+        name = self.element_info.name
+        control_type = self.element_info.control_type
```
vasily-v-ryabov: Invoking button could close the dialog (and crash this code) so it's better to get name and control_type before calling action.

ArtemSkrebkov: Ok. Fixed.

ArtemSkrebkov:
@vasily-v-ryabov 
Please take a look one more time

PR #352, opened 2017-05-09
Split HwndWrapper.send_chars into send_chars and send_keystrokes 

### LINKED ISSUE #334: Method .send_chars() enters every symbol twice for Putty console
This bug is reported in [StackOverflow thread](http://stackoverflow.com/a/42690285/3648361).
Maybe it's a consequence of PR #299.

antonlarin:
There are now 2 methods for silently sending input to the window:
* `send_chars` is supposed to be used to send character input (this includes Enter, Tab, Backspace);
* `send_keystrokes` is for key input (including key combinations with modifiers).

The modifiers in `send_keystrokes` are handled reliably (in my experience with manual testing in different apps) if the target window is inactive but not minimized. For minimized windows behavior varies, e.g.:
* Notepad handles Shift, but not Ctrl or Alt;
* MFC sample CmnCtrl3.exe handles Shift and Ctrl (tried for copy and paste), but not Alt;
* PuTTY handles all modifiers within its terminal emulator, but Alt+F4 fails.

Also, this patch fixes #334.

### DIFF
--- a/pywinauto/controls/hwndwrapper.py
+++ b/pywinauto/controls/hwndwrapper.py
@@ -515,7 +515,9 @@ class HwndWrapper(BaseWrapper):
                              win32con.WA_ACTIVE, 0)
         target_thread_id = user32.GetWindowThreadProcessId(self.handle, None)
         current_thread_id = win32api.GetCurrentThreadId()
-        user32.AttachThreadInput(current_thread_id, target_thread_id, True)
+        attach_success = user32.AttachThreadInput(target_thread_id, current_thread_id, True) != 0
+        if not attach_success:
+            warnings.warn('Failed to attach app\'s thread to the current thread\'s message queue')
 
         keyboard_state_stack = [PBYTE256()]
         user32.GetKeyboardState(ctypes.byref(keyboard_state_stack[-1]))
@@ -523,8 +525,7 @@ class HwndWrapper(BaseWrapper):
         input_locale_id = ctypes.windll.User32.GetKeyboardLayout(0)
         context_code = 0
 
-        keys = keyboard.parse_keys(keystrokes, with_spaces, with_tabs,
-                                   with_newlines)
+        keys = keyboard.parse_keys(keystrokes, with_spaces, with_tabs, with_newlines)
 
         try:
             for key in keys:
@@ -551,10 +552,8 @@ class HwndWrapper(BaseWrapper):
                     new_keyboard_state[vk] |= 128
                     if shift_state & 1 == 1:
                         new_keyboard_state[keyboard.VK_SHIFT] |= 128
-                    if shift_state & 2 == 2:
-                        new_keyboard_state[keyboard.VK_CONTROL] |= 128
-                    if shift_state & 4 == 4:
-                        new_keyboard_state[keyboard.VK_MENU] |= 128
+                    # NOTE: if there are characters with CTRL or ALT in the shift
+                    # state, make sure to add these keys to new_keyboard_state
 
                     keyboard_state_stack.append(new_keyboard_state)
 
@@ -600,7 +599,8 @@ class HwndWrapper(BaseWrapper):
                 warnings.warn(e.strerror)
             user32.SetKeyboardState(ctypes.byref(keyboard_state_stack[0]))
 
-        user32.AttachThreadInput(current_thread_id, target_thread_id, False)
+        if attach_success:
+            user32.AttachThreadInput(target_thread_id, current_thread_id, False)
 
     # -----------------------------------------------------------
     def send_message_timeout(
--- a/pywinauto/controls/hwndwrapper.py
+++ b/pywinauto/controls/hwndwrapper.py
@@ -470,7 +470,6 @@ class HwndWrapper(BaseWrapper):
         If a virtual key with no corresponding character is encountered
         (e.g. VK_LEFT, VK_DELETE), a KeySequenceError is raised.
         """
-
         input_locale_id = ctypes.windll.User32.GetKeyboardLayout(0)
         keys = keyboard.parse_keys(chars, with_spaces, with_tabs, with_newlines)
 
@@ -484,7 +483,7 @@ class HwndWrapper(BaseWrapper):
 
             if unicode_char:
                 _, char = key_info[:2]
-                vk = ctypes.windll.User32.VkKeyScanExW(char, input_locale_id) & 0xFF # TODO: use shift state in high order byte
+                vk = ctypes.windll.User32.VkKeyScanExW(char, input_locale_id) & 0xFF
                 scan = keyboard.MapVirtualKey(vk, 0)
             else:
                 vk, scan = key_info[:2]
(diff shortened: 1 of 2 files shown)


antonlarin:
Additionally, use shift state information from `VkKeyScanEx` to allow characters typed with multiple keys (e.g. `@`) to be passed to `send_keystrokes` verbatim. `send_chars` doesn't need this as characters themselves are sent.

pywinauto/controls/hwndwrapper.py
```diff
@@ -457,52 +460,147 @@ def send_message(self, message, wparam = 0, lparam = 0):
...
+        For more information about Sequences and Modifiers navigate to keyboard.py
+        """
+        user32 = ctypes.windll.User32
+        PBYTE256 = ctypes.c_ubyte * 256
+
+        win32gui.SendMessage(self.handle, win32con.WM_ACTIVATE,
+                             win32con.WA_ACTIVE, 0)
+        target_thread_id = user32.GetWindowThreadProcessId(self.handle, None)
+        current_thread_id = win32api.GetCurrentThreadId()
+        user32.AttachThreadInput(current_thread_id, target_thread_id, True)
```
airelil: A small note, be careful with AttachThreaInput. Msdn says:
A thread cannot attach to itself. Therefore, idAttachTo cannot equal idAttach.

antonlarin: Isn't it impossible for the automated app to be executed in the same thread with pywinauto code? If so, would an assert here suffice?

airelil: It shouldn't, but still, what happens if you wind up with ```self.handle``` be the same as the current thread id? 

Moreover, from MSDN:
The AttachThreadInput function fails if either of the specified threads does not have a message queue. 

I think it's worth checking the functions return statuses.

pywinauto/controls/hwndwrapper.py
```diff
@@ -457,52 +460,147 @@ def send_message(self, message, wparam = 0, lparam = 0):
...
+        Silently send keystrokes to the control
+
+        Parses modifiers Shift(+), Control(^), Menu(%) and Sequences like "{TAB}", "{ENTER}"
+        For more information about Sequences and Modifiers navigate to keyboard.py
+        """
+        user32 = ctypes.windll.User32
+        PBYTE256 = ctypes.c_ubyte * 256
+
+        win32gui.SendMessage(self.handle, win32con.WM_ACTIVATE,
+                             win32con.WA_ACTIVE, 0)
```
vasily-v-ryabov: Why not just `self.set_focus()`? As I remember `WM_ACTIVATE` just requests the window to be in focus if it's possible (blinking sometimes if another app window is active).

pywinauto/controls/hwndwrapper.py
vasily-v-ryabov: Applications like Putty (having console window with no message queues probably) or cmd.exe should fail `AttachThreadInput` call but we want to try sending keystrokes to them anyway. Maybe some warning could be relevant if `AttachThreadInput` was failed, but not a fatal exception.

pywinauto/controls/hwndwrapper.py
antonlarin: `set_focus` steals global keyboard focus and in case of Notepad moves the window to the foreground. As far as I understand, if the user is fine with the window being in the foreground, `type_keys` is the official and the most robust way to pass keys to the application. Therefore, it's important for `send_chars` and `send_keys` to keep the window inactive/minimized.

I've tried multiple ways to make an app respond to silently sending input: doing nothing but attaching a message queue, `WM_ACTIVATE`, `SetActiveWindow`, `SetFocus`, `HwndWrapper.set_focus` and `HwndWrapper.set_keyboard_focus`. For CmnCtrl3.exe these alternatives are seemingly irrelevant (they all work, except for stealing global focus with `HwndWrapper.set_focus` and a crash inside `set_keyboard_focus`). On the other hand, in Notepad the `send_keystrokes` is most functional when activating window via `WM_ACTIVATE` or `SetFocus` WinAPI call.

I understand that this approach to programming is amateurish and the current solution is hacky, but so are both `send_chars` and `send_keystrokes`, because of how inflexible Windows input mechanism is. The best that I can do is to add disclaimers to the docstrings of `send_chars` and `send_keystrokes` explaining their purpose and limitations.

pywinauto/controls/hwndwrapper.py
```diff
@@ -457,52 +460,147 @@ def send_message(self, message, wparam = 0, lparam = 0):
...
-                win32api.SendMessage(self.handle, win32con.WM_KEYDOWN, vk, lparam)
+            else:
+                vk, scan = key_info[:2]
+                char = keyboard.MapVirtualKey(vk, 2)
 
             if char > 0:
                 lparam = 1 << 0 | scan << 16 | (flags & 1) << 24
                 win32api.SendMessage(self.handle, win32con.WM_CHAR, char, lparam)
+            else:
+                raise keyboard.KeySequenceError('no corresponding char for ' +
```
vasily-v-ryabov: Maybe it's more clear to say `'No WM_CHAR code for key {key}, use method send_keystrokes instead'.format(key=key)`

pywinauto/controls/hwndwrapper.py
vasily-v-ryabov: OK, agree with current state of this feature. Thanks for detailed explanation!
Just tested it locally for Notepad and it works great for non-active window. Ctrl- combos don't work for minimized window as expected. OK.

Just think how to warn user that Alt- and Ctrl- won't work for minimized window. Probably we need to check `self.top_level_parent().is_minimized()` and warn user if the sequence contains such a combo.

Sorry for this additional request. But it will really finalize the picture for usability.

pywinauto/controls/hwndwrapper.py
antonlarin: Yes, that would be better, thanks for the suggestion.

pywinauto/controls/hwndwrapper.py
antonlarin: But it may work. Once again, both Ctrl- and Alt- key combinations seem to work for CmnCtrl3.exe, even when it's minimized. Wouldn't the warning be confusing in such a scenario?

Furthermore, when window isn't minimized, but inactive, the method isn't guaranteed to work for every app, because everybody processes input differently. It doesn't work for cmd.exe because it doesn't have an MQ, but I bet there is a GUI app out there that would defy `send_keystrokes`. That's why I'm for putting a disclaimer in the docs. At the end of the day, this method is relatively obscure and the reference will likely be the place where users will find it.

With that being said, if you still think that issuing a warning is better, I'll add it.

vasily-v-ryabov: Yes, it's not guaranteed, but both `send_keystrokes` and `send_chars` work for `cmd.exe` (with warning about `AttachThreadInput`). The following code:
```python
cmd = Application().start("cmd.exe", wait_for_idle=False)
cmd[r'C:\WINDOWS\SYSTEM32\cmd.exe'].send_keystrokes('dir{ENTER}') # or send_chars
```
I think this warning could useful for lazy users. Please add message like `it may not work for minimized window (depends on target app)`.

antonlarin: Oh, sorry, I meant that cmd.exe doesn't see the __modifiers in key combinations__ even when not minimized. Sending `^z` prints just `z` instead of `^z` and so on.

I agree that nobody ever reads the docs and the warning is infinitely better. And I apologize for bargaining, but I think that constraining the warning to minimized windows is wrong if we have on our hands an example of an app (cmd) that works only partially even when the window isn't minimized.

Would you agree to an unconditional warning that says: `send_keystrokes: key combinations may or may not work depending on the target app`?

vasily-v-ryabov: OK, if the sequence contains key combinations, please print warning unconditionally.

PR #355, opened 2017-05-10
Add timeout to connect() with path

daniil-kukushkin:


pywinauto/application.py
```diff
@@ -901,7 +901,7 @@ def connect(self, **kwargs):
             connected = True
 
         elif 'path' in kwargs:
-            self.process = process_from_module(kwargs['path'])
+            self.process = process_from_module(kwargs['path'], timeout=timeout)
```
vasily-v-ryabov: It's better to call `timings.wait_until_passes` or `timings.wait_until` to avoid new param introduction.

pywinauto/application.py
```diff
@@ -912,6 +912,11 @@ def connect(self, **kwargs):
         if not connected:
             raise RuntimeError(
                 "You must specify one of process, handle or path")
+        else:
+            if 'path' not in kwargs:
```
vasily-v-ryabov: `'timeout'` could be in `kwargs` as well (and describe it in the doc string).

pywinauto/application.py
```diff
@@ -912,6 +912,11 @@ def connect(self, **kwargs):
         if not connected:
             raise RuntimeError(
                 "You must specify one of process, handle or path")
+        else:
+            if 'path' not in kwargs:
+                if timeout is not None:
+                    message = "Timeout work only for call connect() with path"
```
vasily-v-ryabov: Should work for process id as well (see `if 'process' in kwargs:`).

pywinauto/unittests/test_application.py
```diff
@@ -270,6 +272,26 @@ def test_connect_path(self):
...
+        """Test that connect_() works with a path with timeout"""
+        app1 = Application()
+        def delayed_launch():
+            time.sleep(2)
+            app1.start(_notepad_exe())
+        thread = Thread(target=delayed_launch)
+        thread.start()
+
+        app_conn = Application()
+        app_conn.connect(path=_notepad_exe(), timeout=3)
```
vasily-v-ryabov: Also need a negative test case (there is no such process, `connect` raises an exception after specified timeout (possible precision is ~0.3 sec).

vasily-v-ryabov: See `self.assertRaises` calls in other tests for reference.

pywinauto/application.py
daniil-kukushkin: @vasily-v-ryabov, I thought about it, but what use case of this behavior? In other words, how to write a unit test, when we do not know the id before the process started?

vasily-v-ryabov: OK, you're right. Skip this.

pywinauto/application.py
```diff
@@ -869,20 +870,26 @@ def __init__(self, backend="win32", datafilename=None):
...
-    def connect(self, timeout=None, **kwargs):
+    def connect(self, **kwargs):
         """Connect to an already running process
 
         The action is performed according to only one of parameters
 
         :param process: a process ID of the target
         :param handle: a window handle of the target
         :param path: a path used to launch the target
+        :param timeout: a timeout for wait process start
```
vasily-v-ryabov: -> `a timeout for process start (relevant if path is specified)`

pywinauto/application.py
```diff
@@ -869,20 +870,26 @@ def __init__(self, backend="win32", datafilename=None):
...
 
         .. seealso::
 
            :func:`pywinauto.findwindows.find_elements` - the keyword arguments that
            are also can be used instead of **process**, **handle** or **path**
         """
+
+        timeout = 0
```
vasily-v-ryabov: It's better to use `None` for default behavior.

pywinauto/application.py
```diff
@@ -901,7 +908,10 @@ def connect(self, timeout=None, **kwargs):
             connected = True
 
         elif 'path' in kwargs:
-            self.process = process_from_module(kwargs['path'], timeout=timeout)
+            if timeout == 0:
+                self.process = process_from_module(kwargs['path'])
+            else:
+                self.process = timings.wait_until_passes(timeout,0,process_from_module,ProcessNotFoundError,kwargs['path'])
```
vasily-v-ryabov: Use PEP8 Luke! `(timeout, 0, process_from_module, ProcessNotFoundError, kwargs['path'])`

pywinauto/application.py
```diff
@@ -913,10 +923,9 @@ def connect(self, timeout=None, **kwargs):
             raise RuntimeError(
                 "You must specify one of process, handle or path")
         else:
-            if 'path' not in kwargs:
-                if timeout is not None:
-                    message = "Timeout work only for call connect() with path"
-                    raise RuntimeError(message)
+            if 'path' not in kwargs and 'timeout' in kwargs:
+                message = "Timeout work only for call connect() with path"
```
vasily-v-ryabov: Grammar: `raise ValueError('Timeout could be specified with path param only')`

daniil-kukushkin:
@vasily-v-ryabov, do you have any ideas about the failed tests?

vasily-v-ryabov:
@daniil-kukushkin we have some unstable tests. Don't worry about that. Everything seems done. Merging.

Issue #347, opened 2017-04-22
Method .wait(...) takes ~ default time (5 sec.) instead of customized timeout like 1 sec.


vasily-v-ryabov:
The bug is reported on StackOverflow:
[pywinauto change default timeout values](http://stackoverflow.com/q/43558087/3648361)

I could reproduce it on simple example using Notepad and `backend='uia'`:
```python
app.window(title_re = 'About Notepad').wait('visible', timeout=1)
```
It took `5.0934726561 sec`.

juysu2:
The default time is for resolve control:
app.window(title_re = 'About Notepad', timeout=1).wait('visible', timeout=1)

vasily-v-ryabov:
The following call took `1.23639702764` sec.
```python
app.window(title_re = 'About Notepad').wait('exists', timeout=1)
```
So it's specific for "visible" state.

vasily-v-ryabov:
I could find the root cause and fixed it! Now it takes around 1 sec:

```python
def t(action):
 	start = time.clock()
 	try:
 		app.window(title_re = 'About Notepad').wait(action, timeout=1)
 	except:
 		print time.clock() - start
>>> t('exists')
1.00364134489
>>> t('visible')
1.09575502439
>>> t('visible')
1.19378547428
>>> t('exists')
1.12031168922
>>> t('exists')
1.03774646544
>>> t('visible')
0.992245803911
>>> t('visible')
1.00437335693
```
Will write unit tests and merge it soon.

PR #357, opened 2017-05-11
Fix wait/wait_not timings, use time.clock(), extend list view detection

vasily-v-ryabov:
* `wait / wait_not` timings are handled properly now (passing params to `_resolve_control`) fixing #347.
* `time.clock()` is resistant to accident system clock change (I found it at the main job recently).
* Extended list view detection list (thanks @KirillMoizik ).


airelil:
@cetygamer is right, it's hard to see the changes in application.py and common_controls.py

vasily-v-ryabov:
It looks like someone of master students used non-default `autoCrLf` value in git config. I'll try to fix it later.
Also I've sent a feedback about GitHub diff viewer that could be improved (`git diff` is not the best option in my mind). Support team replied they will consider it probably.
Possible workaround for now is reviewing every commit except merges.

pywinauto/application.py
```diff
@@ -1,1324 +1,1331 @@
...
+
+        If we already have criteria for both dialog and control then
+        resolve the control and return the requested attribute.
+
+        If we have only criteria for the dialog but the attribute
+        requested is an attribute of DialogWrapper then resolve the
+        dialog and return the requested attribute.
+
+        Otherwise delegate functionality to :func:`__getitem__` - which
+        sets the appropria
```
airelil: A good spot on the explicit call to ```self.__resolve_control``` and timings arguments! I have to admit, I don't understand why ```exists``` needs a special treatment, but if it's the case the ```exists``` processing should be moved a bit upper, before entering into the ```for``` loop, like this: 
```python 
if 'exists' in check_names:
    check = getattr(self, check_name)
    return check(retry_interval, float(retry_interval) // 2)

for check_name in check_names:
    try:
        ...
``` 
The current implementation will demonstrate unstable behavior if ```exists``` will be coupled with any other check.

pywinauto/application.py
```diff
@@ -1044,10 +1044,10 @@ def wait_cpu_usage_lower(self, threshold=2.5, timeout=None, usage_interval=None)
         if timeout is None:
             timeout = Timings.cpu_usage_wait_timeout
 
-        start_time = time.time()
+        start_time = time.clock()
```
cetygamer: Alternatively we could use `timeit` module. It is supposed to be more reliable in terms of calculating elapsed time on different operating systems. See https://docs.python.org/2/library/timeit.html#timeit.default_timer

airelil: I think, ```timeit``` is used mostly for code profiling.

airelil: @vasily-v-ryabov, we may make these time.clock changes later or in a separate PR

pywinauto/application.py
vasily-v-ryabov: @airelil you're partially right. It made me think to write one more unit test to catch the logic fail for case `'exists visible enabled'`. Not sure I will fix it suggested way. But I definitely shouldn't return immediately if `exists` check is True.

pywinauto/application.py
vasily-v-ryabov: OK, will make separate pull requests.

PR #358, opened 2017-05-15
Fix wait/wait_not timings, extend list view detection

vasily-v-ryabov:
* `wait / wait_not` timings are handled properly now (passing params to `_resolve_control`) fixing #347.
* Extended list view detection list (thanks @KirillMoizik ).

@airelil the case with `'exists visible'` for an existing invisible control works somehow so I decided to leave the code as is.


airelil:
@vasily-v-ryabov, and what happens if you switch places: ```visible exists```, does it still work as expected?

vasily-v-ryabov:
Now everything looks correct to me. Any objections?

PR #360, opened 2017-05-21
Make print_control_identifiers faster (fix #359)

### LINKED ISSUE #359: Need to improve print_control_identifiers() performance with depth param
I've observed very slow behavior for iTunes (tab "Store").
```python
>>> len(app.iTunes.descendants()) # takes ~2.5-3 sec.
3024
>>> len(app.iTunes.children()) # takes less than 0.1 sec.
5
```

`app.iTunes.print_control_identifiers()` takes more than a minute! Couldn't wait it for a reasonable time. Even if I set argument `depth=1`.

I think we need to improve this at least for some constrained depth value.

cetygamer:


vasily-v-ryabov:
The result output looks noticeably different. See the attached dump for MS Paint.
[mspaint_controls_new.txt](https://github.com/pywinauto/pywinauto/files/1020192/mspaint_controls_new.txt)
[mspaint_controls_old.txt](https://github.com/pywinauto/pywinauto/files/1020191/mspaint_controls_old.txt)

Didn't measure timings, but functionality for numbering magic attributes like `'TreeUtem1'`, `'TreeUtem2'` etc seems missing with this approach. This was a feature of magic attributes ([rule 3](https://pywinauto.readthedocs.io/en/latest/getting_started.html#how-to-know-magic-attribute-names) and probably others).

Yes, the feature should work when doing `.wrapper_object()` but the idea of `print_control_identifiers()` was to provide pretty the same output as it should work in action.

P.S. We probably need to add more unit tests verifying printed output by comparing vs. a gold file(s).

cetygamer:
So, about an old TODO here:
https://github.com/pywinauto/pywinauto/blob/master/pywinauto/application.py#L581

It is currently impossible to build names for imcomplete controls tree because of the way `findbestmatch.UniqueDict` dinamically adds new names (BaseName0, BaseName1) to already mapped controls.
But then `print_control_identifiers()` works slow because `UiaWrapper.__hash__` returns the same value for a lot of controls.
So I tried to build a map not out of controls themselves but out of their index number in `print_control_identifiers()`.

That did the trick. :relaxed:

@vasily-v-ryabov, please, check if I missed anything this time.

vasily-v-ryabov:
Thanks! Now it looks better than previously. It doesn't assign few numbers to one control.

Was: `[u'22', u'Toolbar12', u'Toolbar10', u'24']`
Became: `[u'22', u'Toolbar10']`

Just need to check if the same approach is applicable for `_resolve_control` method and beyond (search engine itself). Resolving controls and `print_control_identifiers()` must be fully aligned. Otherwise `print_control_identifiers()` doesn't make much sense or may add a confusion.

cetygamer:
Alright, located the issue.

First of all,
> Was: `[u'22', u'Toolbar12', u'Toolbar10', u'24']`
is wrong, there shouldn't be two separate names `ToolbarX` with different numbers. So current (master branch) output of `print_control_identifiers()` is actially incorrect.

The problem was that some controls have really similar properties and in method `__eq__()` of `UIAElementInfo` sometimes gave incorrect results.

vasily-v-ryabov:
Cool! This new implementation of `print_control_identifiers()` takes around 15 seconds for MS Paint example with Open window (`len(dlg.descendants()) == 201` takes ~0.4 sec. for this example).

Current implementation (master branch) takes about 31-33 seconds. So it looks optimized twice. Good! 👍 

Going to merge it but I'm still interesting to optimize it more. With iTunes example (3K controls, 3 sec. for `dlg.descendants()`) printing the whole tree takes more than 10 minutes!

PR #378, opened 2017-06-26
Adds is process running and wait for exit functions

### LINKED ISSUE #376: Wait that process finished
Hi,

What is best way to wait for process to finish with some timeout and if it timeouts then kill it.

I would think something like:
`
timings.wait_until(timeout=10, retry_interval=5, lamda: app.process_exist())
`

efremovd:
Adds two new functions to Application class

is_process_running() -> checks that process is still running
wait_for_process_exit() -> waits for process to finish until timeout (based on timings.wait_until)

Resolves #376


pywinauto/application.py
```diff
@@ -1247,6 +1248,32 @@ def kill(self):
...
+        Returns True if process is still running otherwise - False
+        """
+        is_running = False
+        try:
+            h_process = win32api.OpenProcess(
+                win32con.PROCESS_QUERY_INFORMATION,
+                0,
+                self.process)
+            is_running = win32process.GetExitCodeProcess(
+                h_process) == win32defines.PROCESS_STILL_ACTIVE
```
airelil: What will happen if ```win32api.OpenProcess``` doesn't raise the exception, but h_process is still NULL (check [MSDN](https://msdn.microsoft.com/en-us/library/windows/desktop/ms684320%28v=vs.85%29.aspx?f=255&MSPPError=-2147217396)) ? Will ```win32process.GetExitCodeProcess``` raise the same exception ? I'd add a unit test for it. Leverage mock facilities to return 0 from ```OpenProcess```

pywinauto/application.py
```diff
@@ -1247,6 +1248,32 @@ def kill(self):
...
+            is_running = False
+        return is_running
+
+    def wait_for_process_exit(self, timeout=10, retry_interval=1):
+        """
+        Waits for process to exit until timeout reaches
+
+        Raises TimeoutError exception if timeout was reached
+        """
+        wait_until(timeout, retry_interval, lambda: not self.is_process_running())
```
airelil: How about: 
```wait_until(timeout, retry_interval, self.is_process_running, value=False)``` ?

File: pywinauto/application.py [hunk 3]
```diff
@@ -1247,6 +1248,32 @@ def kill(self):
     # Non PEP-8 aliases
     kill_ = Kill_ = kill
 
+    def is_process_running(self):
+        """
+        Checks that process is still running
+
+        Returns True if process is still running otherwise - False
+        """
```
airelil: Can we run this method before ```application.start``` or ```application.connect``` ?

pywinauto/unittests/test_application.py
```diff
@@ -607,6 +607,26 @@ def test_kill_(self):
...
+        app.start(_notepad_exe())
+        self.assertTrue(app.is_process_running())
+        app.kill()
+        self.assertFalse(app.is_process_running())
+
+    def test_wait_process_exit(self):
+        """Tests wait process to exit"""
+        app = Application()
+        app.start(_notepad_exe())
+        app.UntitledNotepad.Edit.type_keys("hello")
```
airelil: I don't see why we need ```type_keys``` here. I'd simply use ```wait("ready")``` instead.

pywinauto/unittests/test_application.py
```diff
@@ -607,6 +607,26 @@ def test_kill_(self):
...
+        app.UntitledNotepad.menu_select("File->Exit")
+
+        self.assertRaises(TimeoutError, lambda: app.wait_for_process_exit(timeout=5, retry_interval=1))
+        app.Notepad.DontSave.close_click()
+        app.wait_for_process_exit(timeout=10, retry_interval=1)
+        self.assertFalse(app.is_process_running())
+
```
airelil: Again I'd like to see a test with the error handling scenario.

pywinauto/application.py
efremovd: Yes, why not? You meant here to update docstring?

I will add test later for this

pywinauto/application.py
efremovd: Well according to MSDN NULL handle is error result. 

Next GetExitCodeProcess API function will return error saying invalid handle value.

But anyway I can add those tests too

pywinauto/application.py
efremovd: Yep, my bad :)

pywinauto/unittests/test_application.py
efremovd: In next line I'm testing that wait function is timeouted out. I need dialog (Confirmation to save) to popup so wait function will timeout. That's why I'm entering some data to notepad

pywinauto/application.py
efremovd: Added new test with Mocking for OpenProcess function

pywinauto/unittests/test_application.py
airelil: Why there is a need for a dialog, isn't it the same as waiting on the main window?Can't you just call  ```app.wait_for_process_exit```  in ```test_process_is_running``` before calling ```app.kill``` ? In this case, we can remove this test at all.

pywinauto/application.py
```diff
@@ -1247,6 +1248,33 @@ def kill(self):
     # Non PEP-8 aliases
     kill_ = Kill_ = kill
 
+    def is_process_running(self):
+        """
+        Checks that process is running.
+        Can be called before start/connect.
```
vasily-v-ryabov: In this case 3 Codacy warnings are correct. See [Multi-line Docstrings in PEP-257](https://www.python.org/dev/peps/pep-0257/#multi-line-docstrings).

pywinauto/application.py
```diff
@@ -1247,6 +1248,33 @@ def kill(self):
...
+                win32con.PROCESS_QUERY_INFORMATION,
+                0,
+                self.process)
+            is_running = win32process.GetExitCodeProcess(
+                h_process) == win32defines.PROCESS_STILL_ACTIVE
+        except (win32gui.error, TypeError):
+            is_running = False
+        return is_running
+
+    def wait_for_process_exit(self, timeout=10, retry_interval=1):
```
vasily-v-ryabov: Global timeout settings are kept in module `timings`. So I'd suggest to move default values to this module. Also I think too long `retry_interval` is not the best practice because flexible timing assumes it should finish much faster if the condition is already satisfied. So 1 sec. will look too long for user if process is finished after ~0.1 sec. Some reasonable `retry_interval` like 0.01 or 0.1 sec. will not eat too much CPU time but it will look more smoothly.

vasily-v-ryabov:
Thanks for your contribution! I'm planning to make a release on Monday (July, 3). Hope to include these enhancements as well.

PR #429, opened 2017-10-26
Implement scroll method for UIA

### LINKED ISSUE #392: Horizontal scrolling problems (incorrect coordinates for scroll)
The problem is reported and workarounded on StackOverflow:
https://stackoverflow.com/questions/45163039/scrolling-issues-with-pywinauto

airelil:
- Implemented UIAWrapper.scroll - should fix #392 
- Fixed building pattern IDs for 'V2' patterns
- Added UIAWrapper.iface_transform, UIAWrapper.iface_transformV2 and
UIAWrapper.iface_scroll methods

### DIFF
--- a/pywinauto/controls/hwndwrapper.py
+++ b/pywinauto/controls/hwndwrapper.py
@@ -1375,8 +1375,14 @@ class HwndWrapper(BaseWrapper):
             message = win32defines.WM_VSCROLL
 
         # the constant that matches direction, and how much
-        scroll_type = \
-            self._scroll_types[direction.lower()][amount.lower()]
+        try:
+            scroll_type = \
+                self._scroll_types[direction.lower()][amount.lower()]
+        except KeyError:
+            raise ValueError("""Wrong arguments:
+                direction can be any of "up", "down", "left", "right"
+                amount can be any of "line", "page", "end"
+                """)
 
         # Scroll as often as we have been asked to
         if retry_interval is None:
--- a/pywinauto/controls/uiawrapper.py
+++ b/pywinauto/controls/uiawrapper.py
@@ -703,7 +703,6 @@ class UIAWrapper(BaseWrapper):
         **count** (optional) the number of times to scroll
         **retry_interval** (optional) interval between scroll actions
         """
-
         def _raise_attrib_err(details):
             control_type = self.element_info.control_type
             name = self.element_info.name
--- a/pywinauto/controls/uiawrapper.py
+++ b/pywinauto/controls/uiawrapper.py
@@ -36,8 +36,10 @@ from __future__ import print_function
 
 import six
 import comtypes
+import time
 
 from .. import backend
+from ..timings import Timings
 from ..base_wrapper import BaseWrapper
 from ..base_wrapper import BaseMeta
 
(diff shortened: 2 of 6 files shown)


pywinauto/controls/uiawrapper.py
```diff
@@ -652,5 +675,62 @@ def menu_select(self, path, exact=False, ):
...
+        try:
+            scroll_if = self.iface_scroll
+            if direction.lower() in ("up", "down"):
+                if not scroll_if.CurrentVerticallyScrollable:
+                    _raise_attrib_err('is not vertically scrollable')
+            elif direction.lower() in ("left", "right"):
+                if not scroll_if.CurrentHorizontallyScrollable:
+                    _raise_attrib_err('is not horizontally scrollable')
+
+            h, v = self._scroll_types[direction][amount]
```
vasily-v-ryabov: In case of wrong `direction` or `amount` it will raise not very informative `KeyError` for the dict. I'd suggest to raise `ValueError` with possible values listed in the error message.

pywinauto/controls/uiawrapper.py
```diff
@@ -652,5 +675,62 @@ def menu_select(self, path, exact=False, ):
...
+            "page": (uia_defs.scroll_no_amount, uia_defs.scroll_large_increment),
+        },
+    }
+
+    def scroll(self, direction, amount, count=1, retry_interval=None):
+        """Ask the control to scroll itself
+
+        **direction** can be any of "up", "down", "left", "right"
+        **amount** can be one of "line", "page"
+        **count** (optional) the number of times to scroll
```
vasily-v-ryabov: Maybe `(default: 1)` is even more clear as well as for `retry_interval`.

pywinauto/controls/hwndwrapper.py
```diff
@@ -1375,8 +1375,14 @@ def scroll(self, direction, amount, count = 1, retry_interval = None):
...
         # the constant that matches direction, and how much
-        scroll_type = \
-            self._scroll_types[direction.lower()][amount.lower()]
+        try:
+            scroll_type = \
+                self._scroll_types[direction.lower()][amount.lower()]
+        except KeyError:
+            raise ValueError("""Wrong arguments:
+                direction can be any of "up", "down", "left", "right"
+                amount can be any of "line", "page", "end"
```
vasily-v-ryabov: `"end"` is not possible here

airelil: As I can see ```_scroll_types```dictionary does have "end"

vasily-v-ryabov: Maybe it's incomplete push? I've double checked current PR diff and there is no `"end"` key, only `"line"` and `"page"` for every scroll direction.

airelil: I think, there is a bit of confusion here. There is an old, ```Hwndwrapper._scroll_types``` dictionary that does have "end", but ```UIAWrapper._scroll_types``` has been added only in this PR and it indeed doesn't have "end". This is because, win32 back end supports more options than UIA.

vasily-v-ryabov: Ah, got it, thanks!

PR #441, opened 2017-11-12
Work in progress: handling Administrator privileges and other things more gracefully

vasily-v-ryabov:
TODOs:
* Need to handle more Window messages that are not doable with small privileges.
* Need to test at least few such messages (run app as Administrator: `powershell -Command "Start-Process cmd -Verb RunAs"`)
* Need to test access warning while connecting to Admin process.


airelil:
Except of fixing UIA back end. I don't I understand why just verifying privileges at ```connect``` is not enough? The wrapper methods are supposed to work with the same PID, aren't they ?

PR #442, opened 2017-11-19
 Method for get cells in-place-edit wrappers

drinkertea:


pywinauto/controls/common_controls.py
```diff
@@ -657,6 +657,36 @@ def deselect(self):
     # Non PEP-8 alias
     Deselect = deselect
 
+    #-----------------------------------------------------------
+    def search_editor(self, friendly_class_name=""):
```
vasily-v-ryabov: Maybe `inplace_control(...)` is better name.

pywinauto/controls/common_controls.py
```diff
@@ -657,6 +657,36 @@ def deselect(self):
     # Non PEP-8 alias
     Deselect = deselect
 
+    #-----------------------------------------------------------
+    def search_editor(self, friendly_class_name=""):
+        """Return the editor HwndWrapper of the item
+
+        Possible ``friendly_class_name`` values:
+
+        * ``""``  Returns most possible editor of the item
```
vasily-v-ryabov: -> `Return the first appeared in-place control`

pywinauto/controls/common_controls.py
```diff
@@ -657,6 +657,36 @@ def deselect(self):
...
+
+        # Get list of visible controls
+        parent_dlg = self.listview_ctrl.top_level_parent()
+        list_before_click = [w.handle for w in parent_dlg.element_info.descendants() if w.visible]
+
+        # After click in the visible list should editable element appears
+        self.click_input(double=True)
+        list_after_click = [w.handle for w in parent_dlg.element_info.descendants() if w.visible]
+
+        possible_editors = set(list_after_click) - set(list_before_click)
```
vasily-v-ryabov: `possible_inplace_ctrls` is more clear.

pywinauto/controls/common_controls.py
```diff
@@ -657,6 +657,36 @@ def deselect(self):
...
+
+        for handle in possible_editors:
+            hwnd_frclnm = hwndwrapper.HwndWrapper(handle).friendlyclassname
```
vasily-v-ryabov: Please don't use too crazy abbreviations. Let it be `wrapper_fr_class` instead of `hwnd_frclnm`. Also creating the wrapper twice may affect performance (see return statement below).

pywinauto/controls/common_controls.py
```diff
@@ -657,6 +657,36 @@ def deselect(self):
...
+            if (friendly_class_name == "" or hwnd_frclnm == friendly_class_name):
+                return hwndwrapper.HwndWrapper(handle)
+        raise RuntimeError("Element (" + friendly_class_name + ") not found in this list view item")
+    # Non PEP-8 alias
+    SearchEditor = search_editor
```
vasily-v-ryabov: It's a new method. We don't need this alias because there is no point in backward compatibility. :)

pywinauto/controls/common_controls.py
```diff
@@ -657,6 +657,36 @@ def deselect(self):
...
+    def search_editor(self, friendly_class_name=""):
+        """Return the editor HwndWrapper of the item
+
+        Possible ``friendly_class_name`` values:
+
+        * ``""``  Returns most possible editor of the item
+        * ``"friendlyclassname"``  Returns editor with particular friendlyclassname
+        """
+        # If currently editing in this item or some other
+        self.listview_ctrl.type_keys("{ENTER}")
```
vasily-v-ryabov: Probably we need some small pause here while in-place ctrl initialization is done. Or timing inside `type_keys` is enough? I'm not sure. If you decide it, please add new timing in `timings.Timings` object.

pywinauto/controls/common_controls.py
```diff
@@ -657,6 +658,44 @@ def deselect(self):
...
+
+        # Get list of visible controls
+        parent_dlg = self.listview_ctrl.top_level_parent()
+        list_before_click = [w.handle for w in parent_dlg.element_info.descendants() if w.visible]
+
+        # After click in the visible list should editable element appears
```
airelil: "After a click on the visible list an editable element should appear"

pywinauto/controls/common_controls.py
```diff
@@ -657,6 +658,44 @@ def deselect(self):
...
+        self.click_input(double=True)
+        list_after_click = lambda: [w.handle for w in parent_dlg.element_info.descendants() if w.visible]
+
+        try:
+            chek_func = lambda: len(list_after_click()) > len(list_before_click)
```
airelil: Antipattern: assigning lambda expression to a variable.
https://docs.quantifiedcode.com/python-anti-patterns/correctness/assigning_a_lambda_to_a_variable.html

Grammar: check_func

pywinauto/controls/common_controls.py
```diff
@@ -657,6 +658,44 @@ def deselect(self):
...
+
+        Possible ``friendly_class_name`` values:
+
+        * ``""``  Return the first appeared in-place control
+        * ``"friendlyclassname"``  Returns editor with particular friendlyclassname
+        """
+        # If currently editing in this item or some other
+        self.listview_ctrl.type_keys("{ENTER}")
+
+        # Get list of visible controls
```
airelil: "Get a list of visible controls"

drinkertea:
Thx for comments, 12 commits not too much?) i can make new pull from another branch with one commit

airelil:
We can squash and merge. There's no need for another PR

pywinauto/controls/common_controls.py
```diff
@@ -657,6 +658,46 @@ def deselect(self):
...
+    def inplace_control(self, friendly_class_name=""):
+        """Return the editor HwndWrapper of the item
+
+        Possible ``friendly_class_name`` values:
+
+        * ``""``  Return the first appeared in-place control
+        * ``"friendlyclassname"``  Returns editor with particular friendlyclassname
+        """
+        # If currently editing in this item or some other
+        self.listview_ctrl.type_keys("{ENTER}")
```
vasily-v-ryabov: I think this could add potential side effect if we have an editable list box on a dialog with OK and Cancel buttons. This `{ENTER}` can close such a dialog. Maybe another workaround like click to the top-left corner would be better.

pywinauto/controls/common_controls.py
```diff
@@ -657,6 +658,46 @@ def deselect(self):
...
     Deselect = deselect
 
+    #-----------------------------------------------------------
+    def inplace_control(self, friendly_class_name=""):
+        """Return the editor HwndWrapper of the item
+
+        Possible ``friendly_class_name`` values:
+
+        * ``""``  Return the first appeared in-place control
+        * ``"friendlyclassname"``  Returns editor with particular friendlyclassname
```
vasily-v-ryabov: Just an idea for improvement for all controls: related global timings could be documented for every relevant method. It will simplify timings customization.

PR #478, opened 2018-03-29
Update requirements.txt. Add missing dependency.

adolli:
I see you import the module `win32api` in your code https://github.com/pywinauto/pywinauto/blob/fd27458700026f283a23e176bcd7303b68bb5077/pywinauto/mouse.py#L40

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -3,3 +3,4 @@ mock==2.0.0
 rst2pdf
 six==1.10.0
 Sphinx==1.4.1
+pywin32


vasily-v-ryabov:
Thanks for trying. But it looks like a small misunderstanding.

As far as I remember, this `requirements.txt` is used for ReadTheDocs only. It runs on some Linux and pyWin32 is not needed there. Maybe renaming to `doc-requirements.txt` make some sense, but together with ReadTheDocs project change.

`dev-requirements.txt` is used for AppVeyor CI on Windows Server and it contains `pypiwin32` (this is the correct name on PyPI).

Official requirements (for `pip install pywinauto`) are all in `setup.py`. So if you put only `pywinauto` to your local `requirements.txt` it will pull all the dependencies and you don't need to put `comtypes` or `six` there.

adolli:
Sorry for my not considering other platforms. When I try to install pywinauto in virtualenv, the `win32api` is still missing. I see that the dependency is appended dynamically in setup.py. 
https://github.com/pywinauto/pywinauto/blob/master/setup.py#L61-L64

Is there any incompatibility with virtualenv?

vasily-v-ryabov:
Ah, I see. We don't run auto tests in virtualenv. If this is virtualenv specific, it's new finding!

adolli:
I am not sure why it skips installing pypiwin32, not very familiar with virtualenv.

vasily-v-ryabov:
Hmm... I've just figured out that pypiwin32 package became obsolete and pywin32 is now an official PyPI package again. Going to fix it in `setup.py` in next pywinauto update.

adolli:
cool~

PR #485, opened 2018-04-10
Updating Notepad Actions for Windows 10

dibakarbose:
Replaced with "Don't Save" instead of "No"
In Windows 10 version of Notepad option "No" is not present.

So instead of `app.Notepad.No.click()` it's better to use `app.Notepad.DontSave.click()`


vasily-v-ryabov:
Hi @dibakarbose thanks for updating the docs and examples! We appreaciate your help. Just one more improvement could be done: `app.UntitledNotepad` would work while `app.Notepad` doesn't work when English is the default OS language (all the examples rely on that).

vasily-v-ryabov:
Ah, sorry. Just checked and it does work. Merging...

PR #491, opened 2018-04-26
Notepad Automation for Win 10

dibakarbose:
Notepad dialogs and windows style has change in Windows 10. 
Old example works in Windows 7 but not in Windows 10. This file will specifically work on Windows 10 only.
In this program Notepad will be opened and it will first open Page Setup Dialog and after finishing the jobs it will open Print Dialog.

In the end it will write some text in Notepad and then the program will exit.

### DIFF
--- a/examples/notepad_auto_win10.py
+++ b/examples/notepad_auto_win10.py
@@ -1,0 +1,110 @@
+# GUI Application automation and testing library
+# Copyright (C) 2006-2018 Mark Mc Mahon and Contributors
+# https://github.com/pywinauto/pywinauto/graphs/contributors
+# http://pywinauto.readthedocs.io/en/latest/credits.html
+# All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions are met:
+#
+# * Redistributions of source code must retain the above copyright notice, this
+#   list of conditions and the following disclaimer.
+#
+# * Redistributions in binary form must reproduce the above copyright notice,
+#   this list of conditions and the following disclaimer in the documentation
+#   and/or other materials provided with the distribution.
+#
+# * Neither the name of pywinauto nor the names of its
+#   contributors may be used to endorse or promote products derived from
+#   this software without specific prior written permission.
+#
+# THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+# AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+# IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+# DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE
+# FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+# DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+# SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+# CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+# OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+
+#OS Supported : Windows 10 only
+#Goal :: Automating few Notepad operations to
+#For debug purpose I have put some print statements also
+#NOTE : To make the things little slow I have given 1secs of sleep time. You can remove that if you don't want the delay.
+
+
+from pywinauto import application
+import time
+
+app=application.Application()
+app.start("Notepad.exe")
+print("Notepad started")
+
+
+app['Notepad'].wait('ready')
+print("Waiting to be ready")
+
+print("Starting jobs")
+print("Opening Page Setup Dialog")
+app['Notepad'].menu_select("File->PageSetup")
+
+#Selecting Size of the paper
+app["PageSetup"]["Combobox1"].select(2)
+time.sleep(1)
+app.PageSetup.Combobox1.select(1)
+time.sleep(1)
+app.PageSetup.Combobox1.select("Letter")
+time.sleep(1)
+
+print("OK upto this : Size Selection")
+
+#Navigating the radio buttons to select Portrait/Landscape mode.
+app["Page Setup"]["Radio Button1"].click()
+time.sleep(1)
+app["Page Setup"]["Radio Button2"].click()
+time.sleep(1)
+app.PageSetup.RadioButton1.click()
+time.sleep(1)
+print("OK upto this : Orientation Selection")
+
+
+app.PageSetup.Cancel.click()
+#app.PageSetup.OK.click()--> If you want to save the changes
+
+print("OK upto this : Page Setup Dialog")
+
+
+print("Opening Print Dialog")
+app.Notepad.menu_select("File->Print")
+#Selecting printer from the given printer list
+app.print["list item"].select(1)
+time.sleep(1)
+#app.print["list item"].select("Name of the printer exactly as it is") --> You can give the name of the printer as well
+print("OK upto this : Printer selection")
+
+
+#Select preference button
+app.Print.Preferences.click()
+time.sleep(1)
+app["Printing Preferences"].OK.click()
+
+#app.Print["Find Printer..."].click()
+
+app.Print.Cancel.click()
+#app.Print.OK.click() --> If you want to start printing
+
+print("Writing something to Notepad")
+time.sleep(1)
+app['Notepad']['Edit'].set_edit_text("I am typing s\xe4me text to Notepad"
+    "\r\n\r\nAnd then I am going to quit")
+
+print("Exiting Application")
+app.UntitledNotepad.menu_select("File->Exit")
+time.sleep(1)
+app.Notepad.DontSave.click()
+time.sleep(1)
+
+print("Application Closed")


examples/notepad_auto_win10.py
```diff
@@ -0,0 +1,110 @@
...
+# THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+# AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+# IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+# DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE
+# FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+# DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+# SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+# CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+# OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
```
vasily-v-ryabov: I think adding full license test to examples is not needed. If you don't mind, examples could be in public domain. At least I did this for newer examples.

examples/notepad_auto_win10.py
```diff
@@ -0,0 +1,110 @@
...
+
+
+#OS Supported : Windows 10 only
+#Goal :: Automating few Notepad operations to test things
```
vasily-v-ryabov: (minor) As I remember PEP8 requires one space after `#`.

examples/notepad_auto_win10.py
```diff
@@ -0,0 +1,110 @@
...
+#For debug purpose I have put some print statements also
+#NOTE : To make the things little slow I have given 1secs of sleep time. You can remove that if you don't want the delay.
```
vasily-v-ryabov: I would suggest to handle it as it was done in [notepad_slow.py](https://github.com/pywinauto/pywinauto/blob/master/examples/notepad_slow.py):
```python
from pywinauto.timings import Timings
Timings.Slow()
```
`Timings.Slow()` doubles the timings and it could be called few times. Hard-coded `time.sleep`'s are not welcome.

examples/notepad_auto_win10.py
```diff
@@ -0,0 +1,110 @@
...
+from pywinauto import application
+import time
+
+app=application.Application()
+app.start("Notepad.exe")
+print("Notepad started")
+
+
+app['Notepad'].wait('ready')
+print("Waiting to be ready")
```
vasily-v-ryabov: In my mind `Waiting...` means that wait operation is not finished, so it should be before this call, right?

examples/notepad_auto_win10.py
```diff
@@ -0,0 +1,110 @@
...
+
+print("Starting jobs")
+print("Opening Page Setup Dialog")
+app['Notepad'].menu_select("File->PageSetup")
```
vasily-v-ryabov: Some PCs may not have this menu because there is no printer in the system. I think it can be put under condition like `if some_spec.exists():`. This is also a good showcase for method `.exists()`.

dibakarbose:
First of all thanks for reviewing my code. I will make the necessary changes that you requested and commit it.

One thing I would like to ask if ```some_spec.exists()``` is necessary in this case. 
printnotepad

For my system I don't have a printer but it seems some default printer names are already in place(Refer to the screenshot).

I just wanted to clear my doubt that's all

examples/notepad_auto_win10.py
dibakarbose: Yes.. You are right... Sorry for the mistake

examples/notepad_auto_win10.py
dibakarbose: If it is not too much to ask can you mention the differences of using ```Timings.Slow()``` and ```time.sleep()```. Are there any performance issues ?

examples/notepad_auto_win10.py
dibakarbose: Actually I saw in one example that full license has been mentioned. So I thought this is very much necessary.

examples/notepad_auto_win10.py
dibakarbose: I was trying the ```.exists() function``` but somehow I am not able to get the results. I tried the following 
```
>>> app.Print["list item"].exists()
True
```

This works fine.

But I am not able to check for the items in the list.
I tried the following also
```>>> app.Print["list item"]["Microsoft Print to PDF"].exists()```

For this I am getting this error. Maybe I am doing something wrong here. Can you help me out in this.

```
  File "<stdin>", line 1, in <module>
  File "C:\Users\dbosex\AppData\Local\Programs\Python\Python36\lib\site-packages\pywinauto-0.6.4-py3.6.egg\pywinauto\application.py", line 312, in __getitem__
AttributeError: The control does not have a __getitem__ method for item access (i.e. ctrl[key]) so maybe you have requested this in error?
```

Comment (vasily-v-ryabov) [reply examples/notepad_auto_win10.py]: It may be not obvious, but for `.child_window(...)` you have to add `visible_only=False` if it's invisible item. Regarding the last error Win32 backend allows 2 levels of window specification only. For `backend="uia"` the hierarchy is different and more levels are possible.

Comment (vasily-v-ryabov) [reply examples/notepad_auto_win10.py]: So it should be `app.Print["Microsoft Print to PDF"].exists()` in your case. You can check that `app.Print.children()` and `app.Print.descendants()` return the same list of the same length.

**Comment by vasily-v-ryabov** [comment]
My concern about `.exists()` should be handled on a menu item existence level. So it may be before calling Print dialog since it might be unavailable at all. But not sure how to reproduce this situation without removing all virtual printers (which is dangerous of course).

**Comment by dibakarbose** [comment]
@vasily-v-ryabov 

I tried as you suggested. But when I am using ```backend=uia``` I am not able to use ```app.Print["Microsoft Print to PDF"].exists()```

I am not sure about how it is not working. But I found a workaround for this. I don't know whether this is the correct way. The code is working also. Below is the snippet for your reference.

```
app=Application(backend="uia").start('notepad.exe')

app.UntitledNotepad.menu_select("File->Print")

if(app.UntitledNotepad.Print.child_window(title="Microsoft Print to PDF").exists()==True):
    app.UntitledNotepad.Print.child_window(title="Microsoft Print to PDF").select()

```

And one more thing that I observed is that, if I am using ```backend=uia``` then I am not able to use ```app.Notepad.menu_select("File->Print")```. For the code to work properly I have to use ```app.UntitledNotepad.menu_select("File->Print")```

Can you tell me the reason for this behaviour.

**Comment by vasily-v-ryabov** [comment]
`app.Print["Microsoft Print to PDF"].exists()` should work for `backend="win32"`. UIA hierarchy is different and has more levels (just compare `app.top_window().dump_tree()` output for both backends). I'd suggest to concentrate on one backend usage to avoid misunderstandings.

**Comment by dibakarbose** [comment]
Hi,
It has been quite sometime and I haven't updated anything on this. Sorry for this, since I was having some issues and couldn't get time on this.

Will be getting back to this in a few days and do the necessary changes and do a merge request.

Thanks
DIBAKAR

**Comment by vasily-v-ryabov** [comment]
Hi @dibakarbose no problem. Thanks for letting me know. It's easy to update existing pull request by commit/push to the same branch of your fork. PR will be updated automatically. No need to create new PR.

**Comment by dibakarbose** [comment]
> `app.Print["Microsoft Print to PDF"].exists()` should work for `backend="win32"`. UIA hierarchy is different and has more levels (just compare `app.top_window().dump_tree()` output for both backends). I'd suggest to concentrate on one backend usage to avoid misunderstandings.

Hi @vasily-v-ryabov ,  I did as you suggested and tried the following code.

```
>>> from pywinauto import application
>>> app=application.Application(backend="win32").start('notepad.exe')
>>> app.UntitledNotepad.menu_select("File->Print")
>>> app.Print["Microsoft Print to PDF"].exists()
False
```

It seems that whichever backend I am using it is not letting me to check the contents of the list items in the Group Box (Select Printer Box in my case) and hence it is coming as false.

Only for `backend="uia"` the statement 
`app.UntitledNotepad.Print.child_window(title="Microsoft Print to PDF").exists()` is working fine.

One more thing I would like to mention in this:  I used `app.top_window().dump_tree()`  for both backend i.e. "uia" and "win32" and I could find the following lines while using `backend="uia"`

_NOTE :: Following was not available in `backend="win32"`_

```
 |    | ListBox - 'Folder View'    (L573, T187, R969, B273)
   |    | ['Folder View', 'Folder ViewListBox', 'ListBox']
   |    | child_window(title="Folder View", auto_id="1", control_type="List")
   |    |    |
   |    |    | ScrollBar - 'Horizontal'    (L573, T256, R969, B273)
   |    |    | ['ScrollBar', 'Horizontal', 'HorizontalScrollBar', 'ScrollBar0', 'ScrollBar1']
   |    |    | child_window(title="Horizontal", auto_id="NonClientHorizontalScrollBar", control_type="ScrollBar")
   |    |    |    |
   |    |    |    | Button - 'Column left'    (L573, T256, R590, B273)
   |    |    |    | ['Column left', 'Column leftButton', 'Button', 'Button0', 'Button1']
   |    |    |    | child_window(title="Column left", auto_id="UpButton", control_type="Button")
   |    |    |    |
   |    |    |    | Thumb - 'Position'    (L590, T256, R771, B273)
   |    |    |    | ['Position', 'Thumb', 'PositionThumb']
   |    |    |    | child_window(title="Position", auto_id="ScrollbarThumb", control_type="Thumb")
   |    |    |    |
   |    |    |    | Button - 'Page right'    (L771, T256, R952, B273)
   |    |    |    | ['Button2', 'Page rightButton', 'Page right']
   |    |    |    | child_window(title="Page right", auto_id="DownPageButton", control_type="Button")
   |    |    |    |
   |    |    |    | Button - 'Column right'    (L952, T256, R969, B273)
   |    |    |    | ['Column rightButton', 'Button3', 'Column right']
   |    |    |    | child_window(title="Column right", auto_id="DownButton", control_type="Button")
   |    |    |
   |    |    | ListItem - 'Fax'    (L577, T192, R821, B209)
   |    |    | ['FaxListItem', 'Fax', 'ListItem', 'ListItem0', 'ListItem1']
   |    |    | child_window(title="Fax", control_type="ListItem")
   |    |    |
   |    |    | ListItem - 'Microsoft Print to PDF'    (L577, T209, R821, B226)
   |    |    | ['Microsoft Print to PDF', 'Microsoft Print to PDFListItem', 'ListItem2']
   |    |    | child_window(title="Microsoft Print to PDF", control_type="ListItem")
   |    |    |
   |    |    | ListItem - 'Microsoft XPS Document Writer'    (L577, T226, R821, B243)
   |    |    | ['Microsoft XPS Document Writer', 'Microsoft XPS Document WriterListItem', 'ListItem3']
   |    |    | child_window(title="Microsoft XPS Document Writer", control_type="ListItem")
   |    |    |
   |    |    | ListItem - 'Send To OneNote 16'    (L821, T192, R969, B209)
   |    |    | ['Send To OneNote 16ListItem', 'Send To OneNote 16', 'ListItem4']
   |    |    | child_window(title="Send To OneNote 16", control_type="ListItem")
   |    |
   |    | GroupBox - 'Select Printer'    (L563, T170, R980, B339)
   |    | ['GroupBox', 'Select Printer', 'Select PrinterGroupBox', 'GroupBox0', 'GroupBox1']
   |    | child_window(title="Select Printer", auto_id="1072", control_type="Group")
```


So, what should be done now? What is your suggestion in this case?

Is there any other method that I can try?

**Comment by vasily-v-ryabov** [comment]
Well, it seems Printers List is a FolderView element like in `explorer.exe`. It can be automated using `backend="uia"` only. Also I think `backend="uia"` example for Notepad would be something new, because "win32" example already exists.

This line should work (you can check supported patterns for list item in `Inspect.exe`):
```python
app.UntitledNotepad.Print.child_window(title="Microsoft Print to PDF").select()
```

dibakarbose:
> app.UntitledNotepad.Print.child_window(title="Microsoft Print to PDF").select()

I tried the above line and that's working fine.

But now I am stuck with a new issue while trying to select item in ComboBox. I tried to use `Select() `for selecting ComboBox items but it seems that it works only with `backend="win32"`.

Is it not possible to use the same with `backend="uia"` ?

vasily-v-ryabov:
@dibakarbose is it the same as issue #532 ?

dibakarbose:
@vasily-v-ryabov 

I believe it's the same issue.
Below code works fine.
```
>>> from pywinauto import application
>>> app=application.Application(backend="uia").start('notepad.exe')
>>> app.UntitledNotepad.menu_select("File->PageSetup")
>>> app.UntitledNotepad.PageSetup.ComboBox1.exists()
True
```


But this code raises some errors
`>>> app.UntitledNotepad.PageSetup.ComboBox1.select(2)`

First it raises `ValueError: NULL COM pointer access` and along with that it raises another error `NoPatternInterfaceError()`



Using the below code also raises one error
`>>> app.UntitledNotepad.PageSetup.ComboBox1.select('Letter')`

This raises `IndexError("item not found")`
But one observation I could see is that the Combobox is opening and it is trying to select the item. While doing this job it is raising the error

PR #499, opened 2018-05-19
Check admin privileges for both target and Python processes

vasily-v-ryabov:
Alternative implementation might be shorter if checking message in `def send_message` and allow permission for `WM_GETTEXT` only (seems like only this message always works). But it will require using string names for all window messages because there is no guarantee that int message ID is not equal to another attribute of `win32defines`. Example: `self.send_message("CB_GETCOUNT")`. Such changes are as global as current implementation. So don't see a reason to re-implement it.

### DIFF
--- a/pywinauto/application.py
+++ b/pywinauto/application.py
@@ -168,8 +168,8 @@ class WindowSpecification(object):
     def __call__(self, *args, **kwargs):
         """No __call__ so return a usefull error"""
         if "best_match" in self.criteria[-1]:
-            raise AttributeError(
-                "WindowSpecification class has no '{0}' method".
+            raise AttributeError("GUI element (wrapper) is not found " \
+                "or WindowSpecification has no '{0}' method (typo?)".
                 format(self.criteria[-1]['best_match']))
 
         message = (
@@ -956,6 +956,11 @@ class Application(object):
 
         if self.backend.name == 'win32':
             self.__warn_incorrect_bitness()
+
+            if not handleprops.has_enough_privileges(self.process):
+                warning_text = "Python process has no rights to make changes " \
+                    "in the target GUI (run the script as Administrator)"
+                warnings.warn(warning_text, UserWarning)
 
         return self
 
--- a/pywinauto/application.py
+++ b/pywinauto/application.py
@@ -168,8 +168,8 @@ class WindowSpecification(object):
     def __call__(self, *args, **kwargs):
         """No __call__ so return a usefull error"""
         if "best_match" in self.criteria[-1]:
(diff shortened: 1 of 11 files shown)


pywinauto/application.py
```diff
@@ -168,8 +168,8 @@ def __init__(self, search_criteria):
     def __call__(self, *args, **kwargs):
         """No __call__ so return a usefull error"""
         if "best_match" in self.criteria[-1]:
-            raise AttributeError(
-                "WindowSpecification class has no '{0}' method".
+            raise AttributeError("GUI element (wrapper) is not found " \
+                "or WindowSpecification has no '{0}' method (typo?)".
```
cetygamer: I think it is better to highlight the wrong method name.
For example,
```python
"Neither GUI element (wrapper) nor wrapper method '{0}' were found (typo?)"
```

pywinauto/controls/common_controls.py
```diff
@@ -1384,7 +1384,7 @@ def select(self):
             self.elem)                    # item to select
 
         if retval != win32defines.TRUE:
-            raise ctypes.WinError()
+            raise ctypes.WinError()
```
cetygamer: What are the changes up to this point in this file?
Line breaks?

pywinauto/controls/hwndwrapper.py
```diff
@@ -451,10 +451,17 @@ def post_command(self, commandID):
     # Non PEP-8 alias
     #Notify = notify
 
+    # -----------------------------------------------------------
+    def _ensure_enough_privileges(self, message_name):
+        """Ensure the Python process has enough rights to send some window messages"""
+        pid = handleprops.processid(self.handle)
+        if not handleprops.has_enough_privileges(pid):
+            raise OSError('No enough rights to send {} message to target process ' \
```
cetygamer:
```python
"Not enough..."
```

pywinauto/controls/common_controls.py
vasily-v-ryabov: Maybe yes. This is weird, because I have Git option `autocrlf = true` in my settings. I've re-created this PR to avoid much larger diffs for some files. Locally I see only CR-LF line endings everywhere and for these lines particularly. In the repo it should be all `LF`s but it's not easy to check. So we may have the situation when this commit fixes or breaks these `LF`s. I think this is not principal for Windows specific code. Especially because I'm preparing release package on Windows. :)

pywinauto/application.py
vasily-v-ryabov: Agree, will correct with adding some tests.

PR #562, opened 2018-09-05
Add LGTM code quality badges

xcorail:
Hi there!

I thought you might be interested in adding these [code quality badges](https://lgtm.com/projects/g/pywinauto/pywinauto/ci/#badges) to your project. They will indicate a very high code quality to potential users and contributors.
To get an idea of the analyses reflected by these grades, check the alerts discovered by [LGTM](https://lgtm.com/projects/g/pywinauto/pywinauto).

N.B.: I am on the team behind LGTM.com, I'd appreciate your feedback on this initiative, whether you're interested or not, if you find time to drop me a line. Thanks.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,6 +6,8 @@
 [![codecov.io](http://codecov.io/github/pywinauto/pywinauto/coverage.svg?branch=master)](http://codecov.io/github/pywinauto/pywinauto?branch=master)
 [![Code Health](https://landscape.io/github/pywinauto/pywinauto/master/landscape.svg?style=flat)](https://landscape.io/github/pywinauto/pywinauto/master)
 [![Codacy Badge](https://api.codacy.com/project/badge/Grade/b823b31c0f2b48d6873326d038c5a516)](https://www.codacy.com/app/pywinauto/pywinauto?utm_source=github.com&amp;utm_medium=referral&amp;utm_content=pywinauto/pywinauto&amp;utm_campaign=Badge_Grade)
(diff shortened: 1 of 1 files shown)


vasily-v-ryabov:
Hi @xcorail this is interesting project and looks nice. We're using static analyzers for a long time so the more analyzers the better. Some diagnostic messages can be discussed, but there is no 100% ideal static analyzer without false positives. What I could see from current analysis:
 * Few warnings about duplicated `ctypes` import are incorrect because default `import ctypes` doesn't import `ctypes.wintypes` automatically. So we have to do `from ctypes import wintypes` explicitly. I think this moment can be automatically handled for built-in modules at least. The same can be checked for `import pywinauto; from pywinauto import tests` (`__all__` list in `__init__.py` can be checked for this purpose).
 * Unused import under `try-except` with handled `ImportError` is a different case than just simple unused import. I may agree that such a fallback solution can be improved, but some advice for that could be useful. For example trying to `import win32api` can be replaced by `import pkg_resources; pkg_resources.get_distribution('pywin32')` which works in all Python versions but also requires to catch `DistributionNotFound` exception. So there is no much difference between these solutions.
 * It's good you don't warn about "unused imports" in `__init__.py` 👍  which was a false positive in many analyzers previously.
 * Some warnings look really new and useful. I'll consider to use LGTM in our project. Thanks a lot!

xcorail:
Hello @vasily-v-ryabov 

Thanks for this detailed feedback, I really appreciate it!
I will pass the false positives report to my python-analysis colleagues.

Thanks

Issue #503, opened 2018-06-02
Implement key down & key up actions in the keyboard module


vasily-v-ryabov:
This is requested in StackOverflow question: https://stackoverflow.com/questions/50656751/pywinauto-send-key-down-to-application

`class KeyAction` already contains necessary params: `down = True, up = True`. So just need to use them at a higher level.

badari412:
@vasily-v-ryabov Can you please review my PR and let me know if I need to change anything other than the coding standard issues pointed out by codacy?

vasily-v-ryabov:
I think this can be closed after updating the docstring for `keyboard.py`.

vasily-v-ryabov:
Hi @badari412 are you going to update the docs? Or I can do it on my side?

badari412:
@vasily-v-ryabov I just created a pull request related to this. Please look into this.

vasily-v-ryabov:
Fixed in #577 and #584 thanks to @badari412 !

PR #577, opened 2018-10-02
Implement key down & key up actions in the keyboard module #503

badari412:


pywinauto/base_wrapper.py
```diff
@@ -861,31 +861,20 @@ def wait_for_idle(self):
...
+            self,
+            keys,
+            func_to_call,
+            **kwargs):
         self.verify_actionable()
         friendly_class_name = self.friendly_class_name()
+        try:
+            kwargs['pause'] = kwargs['pause'] or Timings.after_sendkeys_key_wait
+        except KeyError:
+            pass
```
vasily-v-ryabov: Empty exception handler is not a good practice. In this case `kwargs.get('pause', Timings.after_sendkeys_key_wait)` looks better. Also I agree to allow `pause=0.0` explicitly (to one's own risk).

pywinauto/base_wrapper.py
```diff
@@ -927,9 +910,47 @@ def type_keys(
...
+          find documentation on what to use for the **keys**.
+
+          .. _keyboard: pywinauto.keyboard.html
+          """
+        return self._type_keys(keys, keyboard.SendKeys, pause=pause, with_spaces=with_spaces, with_tabs=with_tabs,
+                               with_newlines=with_newlines,
+                               turn_off_numlock=turn_off_numlock, set_foreground=set_foreground)
+
+    # -----------------------------------------------------------
+    def key_up(self, key, set_foreground=True):
```
vasily-v-ryabov: I don't think special methods for single key down or up is the only way. Method `type_keys` should provide some notation like `.type_keys("{ALT down}123{ALT up}{SHIFT down}hello{SHIFT up}"`. For single key it may look like `{h down}{i down}{h up}{i up}`.

pywinauto/keyboard.py
```diff
@@ -630,6 +638,20 @@ def SendKeys(keys,
...
-            k.run()
-            time.sleep(pause)
+        _run(keys, pause)
+
+    def _key_up_or_down(key, up, down):
+        keys = parse_keys(key)
+        if isinstance(keys[0].key, six.string_types):
+            keys[0] = EscapedKeyAction(keys[0].key)
+        keys[0].down = down
+        keys[0].up = up
```
vasily-v-ryabov: Since we need marking keys as "up" or "down" in `SendKeys` method, this logic could be moved to `parse_keys`.

pywinauto/keyboard.py
```diff
@@ -605,22 +606,29 @@ def parse_keys(string,
...
     def HiByte(val):
         """Return the high byte of the value"""
         return (val & 0xff00) >> 8
 
+    def _run(keys, pause):
+        for k in keys:
+            k.run()
+            time.sleep(pause)
+
     def SendKeys(keys,
```
vasily-v-ryabov: Since you've started improving this module, can I ask you to rename `SendKeys` to `send_keys`? And make alias `SendKeys = deprecated(send_keys)` where `deprecated` should be imported so: `from . import deprecated`.

pywinauto/base_wrapper.py
badari412: Thanks for your review. Will make changes as you sugested.

pywinauto/base_wrapper.py
badari412: So you want type_keys to handle single key strokes (both keyup and keydown), the way you mentioned above instead of having separate methods?

pywinauto/keyboard.py
badari412: Will do.

pywinauto/base_wrapper.py
vasily-v-ryabov: Yes, I'm not sure about separate methods. Good API should be a little bit redundant, but 10 calls to one method in the test code looks not good.

pywinauto/keyboard.py
badari412: @vasily-v-ryabov I moved the entire logic to handle the combination of keys you mentioned above apart from the existing ones to "parse_keys" function . Please let me know your comments.

```diff
diff --git "a/C:\\Users\\kbad\\AppData\\Local\\Temp\\TortoiseGit\\keyboard-ed91b70.002.py" "b/C:\\MyData\\NonWork\\open-source\\pywinauto\\pywinauto\\keyboard.py"
index 5116f71..9b437b8 100644
--- "a/C:\\Users\\kbad\\AppData\\Local\\Temp\\TortoiseGit\\keyboard-ed91b70.002.py"
+++ "b/C:\\MyData\\pywinauto\\pywinauto\\keyboard.py"
@@ -496,7 +496,7 @@ else:
                 try:
                     pause_time = float(count)
                 except ValueError:
-                    raise KeySequenceError('invalid pause time %s' % count)
+                    raise KeySequenceError('invalid pause time %s'% count)
                 code_keys.append(PauseAction(pause_time))
 
             else:
@@ -524,7 +524,6 @@ else:
 
         return code_keys
 
-
     def parse_keys(string,
                    with_spaces=False,
                    with_tabs=False,
@@ -534,6 +533,8 @@ else:
         keys = []
         if not modifiers:
             modifiers = []
+
+        should_escape_next_keys = False
         index = 0
         while index < len(string):
 
@@ -563,13 +564,27 @@ else:
             # Escape or named key
             elif c == "{":
                 # We start searching from index + 1 to account for the case {}}
-                end_pos = string.find("}", index + 1)
+                end_pos = string.find("}", index+1)
                 if end_pos == -1:
                     raise KeySequenceError('`}` not found')
 
                 code = string[index:end_pos]
                 index = end_pos + 1
-                keys.extend(handle_code(code))
+                key_events = [' up', ' down', ' UP', ' DOWN']
+                current_key_event = None
+                if any(key_event in code for key_event in key_events):
+                    code, current_key_event = code.split(' ')
+                    should_escape_next_keys = True
+                current_keys = handle_code(code)
+                if current_key_event is not None:
+                    if isinstance(current_keys[0].key, six.string_types):
+                        current_keys[0] = EscapedKeyAction(current_keys[0].key)
+
+                    if current_key_event.strip() == 'up':
+                        current_keys[0].down = False
+                    else:
+                        current_keys[0].up = False
+                keys.extend(current_keys)
 
             # unmatched ")"
             elif c == ')':
@@ -596,7 +611,7 @@ else:
                 # if ord(c) in CODE_NAMES:
                 #    keys.append(VirtualKeyAction(ord(c)))
 
-                elif modifiers:
+                elif modifiers or should_escape_next_keys:
                     keys.append(EscapedKeyAction(c))
 
                 else:
@@ -624,10 +639,6 @@ else:
         """Return the high byte of the value"""
         return (val & 0xff00) >> 8
 
-    def _run(keys, pause):
-        for k in keys:
-            k.run()
-            time.sleep(pause)
 
     def SendKeys(keys,
                  pause=0.05,
@@ -638,20 +649,6 @@ else:
         """Parse the keys and type them"""
         keys = parse_keys(keys, with_spaces, with_tabs, with_newlines)
 
-        _run(keys, pause)
-
-    def _key_up_or_down(key, up, down):
-        keys = parse_keys(key)
-        if isinstance(keys[0].key, six.string_types):
-            keys[0] = EscapedKeyAction(keys[0].key)
-        keys[0].down = down
-        keys[0].up = up
-        _run(keys, pause=0.01)
-
-    def key_up(key):
-        """Parses the given key and releases it"""
-        _key_up_or_down(key, up=True, down=False)
-
-    def key_down(key):
-        """Parses the given key and presses and holds it"""
-        _key_up_or_down(key, up=False, down=True)
+        for k in keys:
+            k.run()
+            time.sleep(pause)


```

vasily-v-ryabov: It's easier to leave comment for normal commit than for a diff. The approach seems OK to me. Just need a minor improvement: using str method `.lower()` for checking only `"up"` and `"down"` substrings inclusion.

vasily-v-ryabov: FYI: commit to the same branch will update the PR automatically.

badari412: @vasily-v-ryabov Sorry about it. I didn't want to start the automatic build process until I know it's correct by design. Anyway, I pushed the changes for your review. Thanks!

pywinauto/base_wrapper.py
```diff
@@ -895,7 +906,13 @@ def _type_keys(
             aligned_keys = six.text_type(keys)
 
         # Play the keys to the active window
-        func_to_call(aligned_keys, **kwargs)
+        keyboard.SendKeys(
```
vasily-v-ryabov: Need to call `send_keys` here to avoid DeprecationWarning's for correct `.type_keys` calls.

pywinauto/keyboard.py
```diff
@@ -605,23 +623,26 @@ def parse_keys(string,
...
         return val & 0xff
 
+
     def HiByte(val):
         """Return the high byte of the value"""
         return (val & 0xff00) >> 8
 
-    def SendKeys(keys,
+
+    def send_keys(keys,
```
vasily-v-ryabov: Calls to `SendKeys` should be replaced by `send_keys` also to avoid unnecessary DeprecationWarnings.

vasily-v-ryabov:
Now everything looks good. Just need to update docsting for `keyboard.py` which is automatically published here: https://pywinauto.readthedocs.io/en/latest/code/pywinauto.keyboard.html This can be done in separate PR. I'm merging this one. Thanks a lot!

Issue #399, opened 2017-08-09
top_window() not working when minimized application


blakes22:
Very minor bug (unless it's like that on purpose).
top_window() will not find any window when application is minimized because visible_only argument from find_elements() is False by default - even though process ID and backend are correct.

Current implementation in top_window():

            windows = findwindows.find_elements(process=self.process,
                                                backend=self.backend.name)
   
Works like a charm after adding visible_only=False.

vasily-v-ryabov:
This is design intended because "win32" apps often has several hidden windows that are useless. We can try something like that: (1) use `visible_only=True`; (2) if nothing found, try `visible_only=False`. Does it make sense for you?

vasily-v-ryabov:
One of my thoughts were to remove `top_window()` method at all because it adds a lot of confusion when real top window is changing. It returns WindowSpecification with the handle at some moment and re-usage of this specification confuses a lot.

I think this issue should be solved after we implement an ability to create child window specifications from wrapper object as a parent in search criteria. So that `top_window()` could return a wrapper object.

Another possible way is to have `top_wrapper()` and `top_window()` equivalent to `.window(top_level_only=True, found_index=0)`. But for UIA backend the meaning of top window is different because top window can be a child of main form. So it can be backend specific and therefore removed from `Application` object which is backend agnostic.

PR #580, opened 2018-10-10
Fix for #399-top_window() should work on minimized application

badari412:


pywinauto/application.py
```diff
@@ -1112,14 +1112,19 @@ def top_window(self):
...
             windows = findwindows.find_elements(process=self.process,
-                                                backend=self.backend.name)
+                                                backend=self.backend.name,
+                                                visible_only=only_visible_windows)
+
             if windows:
                 break
             time.sleep(Timings.window_find_retry)
             timeout -= Timings.window_find_retry
+            only_visible_windows ^= True
```
vasily-v-ryabov: Bitwise operation on a bool variable looks a little bit strange. It seems the loop iterations are altered by `True` and `False` values. It looks like a lottery or stochastic algorithm. I'd prefer something more explicit and clear. One of my thoughts were to remove `top_window()` method at all because it adds a lot of confusion when real top window is changing. It returns WindowSpecification with the handle at some moment and re-usage of this specification confuses a lot.

I think this issue should be solved after we implement an ability to create child window specifications from wrapper object as a parent in search criteria. So that `top_window()` could return a wrapper object.

Another possible way is to have `top_wrapper()` and `top_window()` equivalent to `.window(top_level_only=True, found_index=0)`. But for UIA backend the meaning of top window is different because top window can be a child of main form. So it can be backend specific and therefore removed from `Application` object which is backend agnostic.

What do you think?

badari412: As per your comment, I see three possibilities:
1) To remove top_window() 
2) Solve the issue after implementing a feature to create child window specifications using parent in search criteria
3) Make the top_window backend specific and remove it from Application class. 

IMO, decision has to be made based on the amount of time and effort needed. Is it worth the effort? If you ask me I would simply say we get rid of top_window in future releases as I never use top_window() in my tests  for the same reasons you mentioned above.

vasily-v-ryabov: I think point (1) is not good for coming 0.6.6. The second way is the most preferable for me. For the third one it shouldn't be removed from `Application`, but may call or contain backend-specific parts inside. I don't think the third way is worth efforts.

PR #581, opened 2018-10-15
Added method windows() to the class Desktop()

Nebyt:
Added method windows() to the class Desktop().

### DIFF
--- a/pywinauto/__init__.py
+++ b/pywinauto/__init__.py
@@ -109,7 +109,7 @@ if sys.platform == 'win32':
         def __init__(self, backend=None):
             """Create desktop element description"""
             if not backend:
-                raise ValueError('You should set backend, win32 or uia.')
+                self.backend = backends.registry.name
             if backend not in backends.registry.backends:
                 raise ValueError('Backend "{0}" is not registered!'.format(backend))
             self.backend = backends.registry.backends[backend]
--- a/pywinauto/__init__.py
+++ b/pywinauto/__init__.py
@@ -121,7 +121,7 @@ if sys.platform == 'win32':
             if 'backend' in kwargs:
                 raise ValueError('Using another backend than set in Desktop constructor is not allowed!')
             kwargs['backend'] = self.backend.name
-            return WindowSpecification(**kwargs)
+            return WindowSpecification(kwargs)
 
         def windows(self, **kwargs):
             """Return a list of wrapped top level windows"""
--- a/pywinauto/__init__.py
+++ b/pywinauto/__init__.py
@@ -36,6 +36,7 @@ __version__ = "0.6.5"
 
 import sys  # noqa: E402
 import warnings  # noqa: E402
+from .backend import registry
 
 
 def deprecated(method, deprecated_name=None):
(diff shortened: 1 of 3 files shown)


pywinauto/__init__.py
```diff
@@ -104,24 +104,39 @@ def _get_com_threading_mode(module_sys):
 
 
     class Desktop(object):
-
         """Simple class to call something like ``Desktop().WindowName.ControlName.method()``"""
 
-        def __init__(self, backend=None):
+        def __init__(self, backend="win32"):
```
vasily-v-ryabov: I think user must understand explicitly which backend is used. I have an idea to make this argument mandatory for `Application` object as well. But not right now. Maybe in 0.7.0 where we plan to break some compatibility.

pywinauto/__init__.py
```diff
@@ -104,24 +104,39 @@ def _get_com_threading_mode(module_sys):
...
             if 'top_level_only' not in criterion:
                 criterion['top_level_only'] = True
             if 'backend' not in criterion:
-                criterion['backend'] = self.backend
+                criterion['backend'] = self.backend.name
             return WindowSpecification(criterion)
 
+        def windows(self, **criterion):
+            """Return a list of wrapped top level windows"""
+
```
vasily-v-ryabov: Codacy reports this as a code style warning: should be no empty line after a doc string.

pywinauto/unittests/test_application.py
```diff
@@ -1189,22 +1188,41 @@ def test_folder_list(self):
...
 
+        def test_get_list_of_windows_win32(self):
+            """Test that method return list of windows"""
+            dlgs = self.desktop.windows()
+            win_title_result = False
+
+            self.assertTrue(len(dlgs) > 1)
+
+            for win_obj in dlgs:
+                if win_obj.window_text() == self.window_title:
```
vasily-v-ryabov: I think creating a list of window texts makes further test more simple:
```python
window_titles = [win_obj.window_text() for win_obj in dlgs]
self.assertTrue(self.window_title in window_titles)
```

pywinauto/__init__.py
```diff
@@ -104,24 +104,39 @@ def _get_com_threading_mode(module_sys):
...
         def window(self, **criterion):
             """Create WindowSpecification object for top-level window"""
             if 'top_level_only' not in criterion:
                 criterion['top_level_only'] = True
             if 'backend' not in criterion:
-                criterion['backend'] = self.backend
+                criterion['backend'] = self.backend.name
             return WindowSpecification(criterion)
 
+        def windows(self, **criterion):
```
vasily-v-ryabov: I'd prefer `**kwargs` name to be consistent with `Application`.

pywinauto/__init__.py
```diff
@@ -104,24 +104,39 @@ def _get_com_threading_mode(module_sys):
...
+            """Return a list of wrapped top level windows"""
+
+            if 'backend' in criterion:
+                raise ValueError('Do not override backend!')
+
+            if 'visible_only' not in criterion:
```
vasily-v-ryabov: Maybe conditions could be covered by auto tests better. I mean covering exception and `visible_only`/`enabled_only` in `**kwargs`.

pywinauto/unittests/test_application.py
Nebyt: Okay, i will change this moment.

pywinauto/__init__.py
Nebyt: I will return `None` for backend default value and will add a check for `None` value.

Nebyt: Something like this:
```python
if not backend:
    raise ValueError('You should set backend, win32 or uia.')
```

pywinauto/__init__.py
Nebyt: I think I undestand how can i do this check.

pywinauto/unittests/test_application.py
```diff
@@ -1189,22 +1188,70 @@ def test_folder_list(self):
...
+            """Test that method return list of windows"""
+            dlgs = self.desktop.windows()
+            self.assertTrue(len(dlgs) > 1)
+
+        def test_set_backend_to_windows_uia(self):
+            """Set backend to method windows, except exception ValueError"""
+            with self.assertRaises(ValueError):
+                self.desktop.windows(backend='win32')
+            with self.assertRaises(ValueError):
+                self.desktop.windows(backend='uia')
```
vasily-v-ryabov: I know this is not a subject, but similar checks are useful for method `window()` as well.

pywinauto/unittests/test_application.py
```diff
@@ -1189,22 +1188,70 @@ def test_folder_list(self):
                               u'CtrlTest.exe', u'mfc100u.dll', u'RebarTest.exe', u'RowList.exe', u'TrayMenu.exe'])
             self.assertEqual(files_list.item('RebarTest.exe').window_text(), 'RebarTest.exe')
 
+        def test_get_list_of_windows_uia(self):
+            """Test that method return list of windows"""
```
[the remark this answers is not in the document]
Nebyt: I did change.

pywinauto/unittests/test_application.py
Nebyt: Okay, I will add it to the method window() too.

pywinauto/__init__.py
```diff
@@ -104,23 +104,39 @@ def _get_com_threading_mode(module_sys):
...
-            if 'top_level_only' not in criterion:
-                criterion['top_level_only'] = True
-            if 'backend' not in criterion:
-                criterion['backend'] = self.backend
-            return WindowSpecification(criterion)
+            if 'top_level_only' not in kwargs:
+                kwargs['top_level_only'] = True
+            if 'backend' not in kwargs:
+                kwargs['backend'] = self.backend.name
+            return WindowSpecification(kwargs)
```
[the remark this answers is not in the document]
Nebyt: Hm... Are you sure it's necessary? Because in the Application.window() we have this:
```python
    win_spec = WindowSpecification(kwargs)
return win_spec
```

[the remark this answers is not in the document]
Nebyt: Ok you are right). I check it out in IDLE, will fix this in Desktop.window() and Application.window().

PR #624, opened 2018-12-08
Epic cleanup of deprecation warnings in unittests, split two

airelil:


pywinauto/controlproperties.py
```diff
@@ -80,11 +80,15 @@ def window_text(self):
...
+    def has_style(self, style):
         return self['style'] & style == style
+    # Non PEP-8 alias
+    HasStyle = deprecated(has_style)
 
-    def HasExStyle(self, exstyle):
+    def has_exstyle(self, exstyle):
         return self['exstyle'] & exstyle == exstyle
+    # Non PEP-8 alias
+    HasExStyle = deprecated(has_exstyle)
```
vasily-v-ryabov: It should be `HasExStyle = deprecated(has_exstyle, deprecated_name='HasExStyle')` because it has non-standard conversion. Standard name conversion warns to replace "has_exstyle" by "HasExstyle". There are very few corner cases like that so this optional parameter has been added to point user in the right direction.

vasily-v-ryabov: Sorry, it warns to replace "HasExStyle" by "has_ex_style". But the method is named `has_exstyle`.

PR #655, opened 2019-01-13
Fix #647 (decode error while listing menu items for "win32" backend)

### LINKED ISSUE #647: Unable to select menu item from popup menu in Sublime Text
I can automate right click on an empty file opened on sublime. But once I get the menu from the right click I cannot seem to select any of the options like `Paste` or `Open containing folder` shown on the menu n Sublime.Any way I can do that? 

Need access/focus/control of that popup menu before I choose the option I want to click. Struggling to achieve that.  Any help be appreciated. Thanks.

vasily-v-ryabov:
subj

### DIFF
--- a/pywinauto/controls/menuwrapper.py
+++ b/pywinauto/controls/menuwrapper.py
@@ -367,7 +367,7 @@ class MenuItem(object):
         if six.PY3:
             return "<MenuItem " + self.text() + ">"
         else:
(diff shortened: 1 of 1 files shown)


airelil:
We may need to make similar changes in other __repr__. For example, in uia_element_info

vasily-v-ryabov:
`uia_element_info` requires separate `__repr__` implementation, it's inherited from base class now and seems extremely rarely used. We may submit low priority issue on that.

Issue #691, opened 2019-03-21
ctypes.ArgumentError: argument 1: <class 'OverflowError'>: int too long to convert


frogger72:
Getting the error when calling `is_minimized()`
Using pywinauto 0.6.6 on python 3.6.8 64-bit (Intel channel) to automate 64-bit Excel 2016

```python
Traceback (most recent call last):
  File "C:\proj\proj.py", line 479, in func
    if wnd.is_minimized():
  File "C:\Miniconda3\envs\py3\lib\site-packages\pywinauto\application.py", line 378, in __getattribute__
    ctrls = self.__resolve_control(self.criteria)
  File "C:\Miniconda3\envs\py3\lib\site-packages\pywinauto\application.py", line 256, in __resolve_control
    criteria)
  File "C:\Miniconda3\envs\py3\lib\site-packages\pywinauto\timings.py", line 436, in wait_until_passes
    func_val = func(*args, **kwargs)
  File "C:\Miniconda3\envs\py3\lib\site-packages\pywinauto\application.py", line 201, in __get_ctrl
    dialog = self.backend.generic_wrapper_class(findwindows.find_element(**criteria[0]))
  File "C:\Miniconda3\envs\py3\lib\site-packages\pywinauto\findwindows.py", line 84, in find_element
    elements = find_elements(**kwargs)
  File "C:\Miniconda3\envs\py3\lib\site-packages\pywinauto\findwindows.py", line 281, in find_elements
    elements = [elem for elem in elements if _title_match(elem)]
  File "C:\Miniconda3\envs\py3\lib\site-packages\pywinauto\findwindows.py", line 281, in <listcomp>
    elements = [elem for elem in elements if _title_match(elem)]
  File "C:\Miniconda3\envs\py3\lib\site-packages\pywinauto\findwindows.py", line 277, in _title_match
    t = w.rich_text
  File "C:\Miniconda3\envs\py3\lib\site-packages\pywinauto\win32_element_info.py", line 83, in rich_text
    return handleprops.text(self.handle)
  File "C:\Miniconda3\envs\py3\lib\site-packages\pywinauto\handleprops.py", line 91, in text
    handle, win32defines.WM_GETTEXT, length, ctypes.byref(buffer_))
ctypes.ArgumentError: argument 1: <class 'OverflowError'>: int too long to convert
```

airelil:
Seems to be the same as #641 . Needs additional cleaning in win32functions.py

[...]

Issue #641, opened 2018-12-27
win32functions.IsWindowVisible issue: OverflowError: int too long to convert


vasily-v-ryabov:
The issue has been recently reported on StackOverflow: https://stackoverflow.com/q/53942453/3648361

airelil:
Oh, forgot to mention, and we can set IsWindowVisible.argtypes as well

vasily-v-ryabov:
Yes, this is for `argtypes` change, of course.

Issue #650, opened 2019-01-07
Use wintypes in win32functions.py and win32_structure.py wherever possible


vasily-v-ryabov:
It's usually more accurate and different for 32/64 bit. For example, `LPARAM` and `WPARAM` definitions.

vasily-v-ryabov:
@airelil please close if we have finished with this.

PR #692, opened 2019-03-23
Issue #691: tighten args for SendMessage and GetWindowLong

airelil:
Another fix for win32 signatures. Similarly to issue #641 it should help to process window handles with values beyond usual int range.

The last arg for SendMessage, actually, should be LPARAM, but it's tricky to pass it as a pointer from ctypes.byref() so LPVOID is selected instead

Also fixed a typo for GetClassName return type and PEP-8 typos

### DIFF
--- a/pywinauto/handleprops.py
+++ b/pywinauto/handleprops.py
@@ -219,6 +219,8 @@ def rectangle(handle):
 def font(handle):
     """Return the font as a LOGFONTW of the window"""
     # get the font handle
+    if handle is None:
+        handle = 0  # make sure we don't pass window handle down as None
     font_handle = win32functions.SendMessage(
         handle, win32defines.WM_GETFONT, 0, 0)
 
--- a/pywinauto/handleprops.py
+++ b/pywinauto/handleprops.py
@@ -145,13 +145,13 @@ def contexthelpid(handle):
 #=========================================================================
 def iswindow(handle):
     """Return True if the handle is a window"""
-    return False if handle is None else  bool(win32functions.IsWindow(handle))
+    return False if handle is None else bool(win32functions.IsWindow(handle))
 
 
 #=========================================================================
 def isvisible(handle):
     """Return True if the window is visible"""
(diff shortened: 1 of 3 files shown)


pywinauto/win32functions.py
```diff
@@ -103,13 +103,15 @@
 ShowWindow			= 	ctypes.windll.user32.ShowWindow
 GetWindowContextHelpId =	ctypes.windll.user32.GetWindowContextHelpId
 GetWindowLong		=	ctypes.windll.user32.GetWindowLongW
+GetWindowLong.argtypes = [win32structures.HWND, ctypes.c_int]
```
vasily-v-ryabov: We can rely on built-in Python definitions in `ctypes.wintypes` for some types:
```python
from ctypes import wintypes
wintypes.HWND
```

pywinauto/handleprops.py
```diff
@@ -145,13 +145,13 @@ def contexthelpid(handle):
...
     """Return True if the handle is a window"""
-    return False if handle is None else  bool(win32functions.IsWindow(handle))
+    return False if handle is None else bool(win32functions.IsWindow(handle))
 
 
 #=========================================================================
 def isvisible(handle):
     """Return True if the window is visible"""
-    return False if handle is None else  bool(win32functions.IsWindowVisible(handle))
+    return False if handle is None else bool(win32functions.IsWindowVisible(handle))
```
vasily-v-ryabov: Maybe `win32gui.IsWindowVisible` can be used. But I'm not sure if it can raise some exception which we don't handle here.

pywinauto/win32functions.py
airelil: Meanwhile I just follow the same convention as the rest of the signatures in this module. Let's keep it consistent. I think, we already talked that we'll have a separate commit to switch to definitions from ctypes

vasily-v-ryabov: OK, we can fix it separately. I've found it's issue #650.

PR #724, opened 2019-04-28
[Issue#627] Add is_editable method for editwrapper

drinkertea:


pywinauto/unittests/test_uiawrapper.py
```diff
@@ -1293,6 +1293,31 @@ def tearDown(self):
...
+
+        def test_edit_editable(self):
+            """Test editable method for editable edit"""
+            self.assertEqual(self.dlg.Edit2.get_value(), "Editable")
+            self.assertTrue(self.dlg.Edit2.is_editable())
+
+        def test_edit_readonly(self):
+            """Test editable method for read only edit"""
+            self.assertEqual(self.dlg.Edit1.get_value(), "ReadOnly")
+            self.assertFalse(self.dlg.Edit1.is_editable())
```
airelil: I would just make one test instead of these two. It woull save a bit running time for the whole test matrix.
```python
        def test_is_editable(self):
            """Test is_editable method for Edit control"""
            self.assertEqual(self.dlg.Edit2.get_value(), "Editable")
            self.assertTrue(self.dlg.Edit2.is_editable())
            self.assertEqual(self.dlg.Edit1.get_value(), "ReadOnly")
            self.assertFalse(self.dlg.Edit1.is_editable())
```

pywinauto/unittests/test_uiawrapper.py
```diff
@@ -1293,6 +1293,31 @@ def tearDown(self):
...
+
+        def tearDown(self):
```
airelil: Usually ```tearDown``` follows after ```setUp``` and before the tests...

Issue #419, opened 2017-10-05
ctypes.ArgumentError @ click_input


Enteleform:
> Windows 10.0.15063 x64
> Python 3.6.2
> pywinauto 0.6.3

&nbsp;  
Setup (functional):
```python
hwnd = my_kivy_app.get_hwnd()

app = pywinauto.Application()
app.connect(handle=hwnd)
window = app.window(handle=hwnd).wrapper_object()
```

&nbsp;  
Exception @:
```python
window.click_input(button='left', pressed='', coords=(100, 100), double=False, absolute=False)
```

```java
Traceback (most recent call last):
  File "Test_Window_Commands.py", line 114, in <module>
    app.run()
  File "C:\OneDrive\_Frameworks\Python\3\lib\site-packages\kivy\app.py", line 828, in run
    runTouchApp()
  File "C:\OneDrive\_Frameworks\Python\3\lib\site-packages\kivy\base.py", line 504, in runTouchApp
    EventLoop.window.mainloop()
  File "C:\OneDrive\_Frameworks\Python\3\lib\site-packages\kivy\core\window\window_sdl2.py", line 663, in mainloop
    self._mainloop()
  File "C:\OneDrive\_Frameworks\Python\3\lib\site-packages\kivy\core\window\window_sdl2.py", line 405, in _mainloop
    EventLoop.idle()
  File "C:\OneDrive\_Frameworks\Python\3\lib\site-packages\kivy\base.py", line 339, in idle
    Clock.tick()
  File "C:\OneDrive\_Frameworks\Python\3\lib\site-packages\kivy\clock.py", line 581, in tick
    self._process_events()
  File "kivy\_clock.pyx", line 367, in kivy._clock.CyClockBase._process_events (kivy\_clock.c:7700)
  File "kivy\_clock.pyx", line 397, in kivy._clock.CyClockBase._process_events (kivy\_clock.c:7577)
  File "kivy\_clock.pyx", line 395, in kivy._clock.CyClockBase._process_events (kivy\_clock.c:7498)
  File "kivy\_clock.pyx", line 167, in kivy._clock.ClockEvent.tick (kivy\_clock.c:3483)
  File "C:\OneDrive\_Projects\Python\dev\kivy\basic\_app.py", line 140, in _wrapped
    function(*args, **kwargs)
  File "Test_Window_Commands.py", line 104, in ctrl_shift_alt
[...]

PR #732, opened 2019-05-15
ctypes adaptations in order to work with other packages

CristiFati:
This is related to [\[GitHub\]: ctypes.ArgumentError @ click\_input](https://github.com/pywinauto/pywinauto/issues/419) (my comment suggesting extending classes from *wintypes* instead of redefining them).

As a test example, I used the same code from [\[SO\]: ctypes.ArgumentError when using kivy with pywinauto (@CristiFati's answer)](https://stackoverflow.com/questions/55928463/ctypes-argumenterror-when-using-kivy-with-pywinauto/55931295#55931295).

Of course, switching back to *ctypes*, brought everything back. I successfully tested but changes were needed on *kivy*'s side as well. Once this *PR* passes (if it does), I will submit another one to them.

### DIFF
--- a/pywinauto/win32structures.py
+++ b/pywinauto/win32structures.py
@@ -153,13 +153,9 @@ LPARAM = wintypes.LPARAM
 WPARAM = wintypes.WPARAM
 
 
-class POINT(Structure):
-    _pack_ = 4
-    _fields_ = [
-        # C:/PROGRA~1/MIAF9D~1/VC98/Include/windef.h 307
-        ('x', LONG),
-        ('y', LONG),
-    ]
+class POINT(wintypes.POINT):
+
+    """Wrap the POINT structure and add extra functionality"""
 
     def __iter__(self):
         """Allow iteration through coordinates"""
--- a/pywinauto/win32structures.py
+++ b/pywinauto/win32structures.py
@@ -153,7 +153,7 @@ LPARAM = wintypes.LPARAM
 WPARAM = wintypes.WPARAM
 
 
-class POINT(wintypes.POINT, Structure):
+class POINT(wintypes.POINT):
 
     """Wrap the POINT structure and add extra functionality"""
 
@@ -170,6 +170,9 @@ class POINT(wintypes.POINT, Structure):
             return self.y
         else:
             raise IndexError("Illegal index")
+
+    __eq__ = Structure.__eq__
+    __str__ = Structure.__str__
 
 assert sizeof(POINT) == 8, sizeof(POINT)
 assert alignment(POINT) == 4, alignment(POINT)
--- a/pywinauto/win32structures.py
+++ b/pywinauto/win32structures.py
@@ -41,9 +41,9 @@ from .win32defines import LF_FACESIZE
 from . import sysinfo
 
 
-class Structure(Struct):
+class StructureMixIn:
 
-    """Override the Structure class from ctypes to add printing and comparison"""
+    """Define printing and comparison behaviors to be used for the Structure class from ctypes"""
 
     #----------------------------------------------------------------
     def __str__(self):
@@ -91,6 +91,11 @@ class Structure(Struct):
                 return False
 
         return False
+
+
+class Structure(Struct, StructureMixIn):
+    """Override the Structure class from ctypes to add printing and comparison"""
+    pass
 
 ##====================================================================
 #def PrintCtypesStruct(struct, exceptList = []):
@@ -153,7 +158,7 @@ LPARAM = wintypes.LPARAM
 WPARAM = wintypes.WPARAM
 
 
-class POINT(wintypes.POINT):
+class POINT(wintypes.POINT, StructureMixIn):
 
     """Wrap the POINT structure and add extra functionality"""
 
@@ -171,8 +176,6 @@ class POINT(wintypes.POINT):
         else:
             raise IndexError("Illegal index")
 
-    __eq__ = Structure.__eq__
-    __str__ = Structure.__str__
 
 assert sizeof(POINT) == 8, sizeof(POINT)
 assert alignment(POINT) == 4, alignment(POINT)
--- a/pywinauto/win32structures.py
+++ b/pywinauto/win32structures.py
@@ -99,6 +99,7 @@ class Structure(Struct, StructureMixIn):
 
     pass
 
+
 ##====================================================================
 #def PrintCtypesStruct(struct, exceptList = []):
 #    """Print out the fields of the ctypes Structure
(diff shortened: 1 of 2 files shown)


pywinauto/win32structures.py
```diff
@@ -153,13 +153,9 @@ def _reduce(self):
...
 
 
-class POINT(Structure):
-    _pack_ = 4
-    _fields_ = [
-        # C:/PROGRA~1/MIAF9D~1/VC98/Include/windef.h 307
-        ('x', LONG),
-        ('y', LONG),
-    ]
+class POINT(wintypes.POINT, Structure):
```
vasily-v-ryabov: Is `wintypes.POINT` not enough? It's not clear why multiple inheritance is needed here.

CristiFati: @vasily-v-ryabov: The unit tests failed (this is why I added this in my 3rd commit). By default `ctypes.Structure` descendents don't know how to compare their *\_fields\_* content to a tuple, that's what `Structure.__eq__` does. I wanted to use the mixin technique, but apparently I introduced a nice diamond of hell. As soon as tests finish, I'll correct the issue (replacing it by a workaround).

Secondly, I don't know why I get coverage reports stating a drop from 90+% to 60+%.

airelil: Regarding the coverage report, I think, it's about changes of signatures for functions that are never called.

CristiFati:
Thank you both for the comments. It seems like the coverage is "mysteriously" back to normal. I assume (unless any of you did some tricks behind the scenes), that it dropped because of the filed test(s), all the test scheduled after it were no longer run, hence the lower coverage percent (I didn't check times needed to run to see if they match the percentages).

Regarding the filed tests (I don't know why the one before the last failed). I noticed this inconsistent behavior on my machine too. I have some questions:

- Is it possible once a test failed to cancel all the remaining ones (as *AppVeyor* take long time)
- Is it possible to trigger tests without a (dummy) commit?
- Is it possible trigger tests (not part of a *PR*), as I could have submitted the *PR* when I was in this state (to avoid all the "public" back and forth transitions)?

pywinauto/win32structures.py
vasily-v-ryabov: Coverage report starts from 60% because every Python version submits coverage and the bot posts incomplete coverage first, then it edits the post with every update after every line of AppVeyor matrix done.

Regarding commit email you can see that all commits are gray and not bound to your GitHub account. So your contribution won't be counted correctly here: https://github.com/pywinauto/pywinauto/graphs/contributors

The reason of the gray commits that you used Git local settings for "username" and "email" different than in your GitHub account. See https://help.github.com/en/articles/setting-your-commit-email-address-in-git

vasily-v-ryabov:
You're free to create AppVeyor project for your forked repo and trigger tests manually (without a commit) or by commit to any branch without creating a PR. Just customize the repo settings (webhooks) and AppVeyor project settings for your needs. Feel free to ask us if some steps are not clear.

Creating/updating PR in the main repo triggers AppVeyor job in any case. Only maintainers can cancel or re-run jobs in this AppVeyor project.

pywinauto/win32structures.py
CristiFati: @vasily-v-ryabov: Thank you for the details.

Regarding the username, I figured that out in the meantime and I changed it (by default it was empty, and my PC username was used instead), but apparently for pushed commits it's not a trivial task. So I don't mind if my username won;t appear in the contributors list. My goal is having this fixed, not going into the hall of fame.

pywinauto/win32structures.py
```diff
@@ -41,9 +41,9 @@
 from . import sysinfo
 
 
-class Structure(Struct):
+class StructureMixIn:
```
vasily-v-ryabov: Every class should be inherited at least from `object` to be "new-style class" in Python 2.7.

pywinauto/win32structures.py
vasily-v-ryabov: No problem. I can fix this if you send me an email to `"vasily-v-ryabov".replace('-', '.')` at gmail(dot)com. `git rebase` command is needed. Maybe this script is useful: https://help.github.com/en/articles/changing-author-info (but I'd like to check this carefully).

vasily-v-ryabov: One more useful answer: https://stackoverflow.com/a/1320317/3648361

CristiFati:
@vasily-v-ryabov: Thank you for the git related tips. Many things are new to me :)

CristiFati:
1. All, although *Pywinauto* is more correct due to this refactor, in order for the end user to see the effects, *Kivy* also needs to be updated. That's why I submitted [\[GitHub\]: kivy/kivy - ctypes adaptations in order to work with other packages](https://github.com/kivy/kivy/pull/6319).

2. As you probably noticed, I'm not too much of an expert in *git* / *GitHub*.  After my merge, if I want to merge another fix, I have to sync my online repo (*CristiFati/pywinauto*) from the official one (*pywinauto/pywinauto*) first. I'm following the steps from [\[GitHub.Help\]: Syncing a fork](https://help.github.com/en/articles/syncing-a-fork). <br>The problem is that at step ***5.*** (*merge*), an additional commit ("Merged upstream....") is automatically created, and as a consequence pushed onto my online repo, and therefore it will be included in the next *PR*, which is obviously something that I'd avoid. I looked at your (online) repos and I didn't see anything similar. So what am I missing? Should I *rebase* instead of *merge*? <br>An alternative that works, but I'd avoid, is *delete* + *re-fork* :)

@airelil:
I don't know what the "*LGTM alerts*" alerts are, but I looked a bit into your repo, and I saw an implementation of *\_\_hash\_\_* relying on *id*, which will lead to cases when 2 objects that test equal, will have different hashes - which doesn't make too much sense. Probably this is what you meant

vasily-v-ryabov:
Rebase from upstream is what you need to avoid redundant merge commit: https://gist.github.com/ravibhure/a7e0918ff4937c9ea1c456698dcd58aa
It's much easier than re-creating fork that requires re-setup of AppVeyor hooks etc.

BTW, we're thinking about re-writing [py_inspect](https://github.com/pywinauto/py_inspect) (written in PyQt5) using Kivy. Does it make sense in terms of minimizing executable package size or other benefits? Since you have more experience with Kivy, your help would be valuable.

CristiFati:
@vasily-v-ryabov: Thank you for the answer! Regarding your question, my interaction with *Kivy* hardly qualifies as experience, as I noticed the *SO* question, installed the 2 modules, copy / paste an example, then a few hours browsing through (a very specific area of) the code and that's it. I am not really a *GUI* guy. I don't have too much experience of any of the modules you mentioned (for example my main *VEnv* has *2+GiB*, as I install modules and play with them for a little while - but that's only scratching the surface), so I'm afraid my opinion isn't very relevant here :(

vasily-v-ryabov:
OK, I've asked just in case. Thanks for your efforts.

Issue #706, opened 2019-04-08
find_best_control_matches causes ComboBox flutter


bernd-wechner:
This is a tad bothersome to say the least. When i try to find a control by its name like:

`control = win["name"]`

all the ComboBoxes on my form flutter. It took me some serious drill down effort to nail this to:

`ComboBoxWrapper.texts`

which literally expands the combobox and collapses it again all in the name, it seems of finding three child controls that turn out to be:
```
uia_controls.ListViewWrapper - '', ListBox
uia_controls.StaticWrapper - '', Static
uia_controls.ButtonWrapper - 'Close', Button
```

But I'm looking for a panel in fact (Pane10), and I find it and it's all good and the software alas has no title or name on that control, and so forth, but in finding it I have to endure unwanted ComboBox flutter?

What is the work around here? I'm curious. I will explore it some more, but I really don't need all the ComboBox children to be checked when looking for a control unless I'm looking for one of those children. In short I kind of expect and guess this is a hope for, a setting through which I can say: Don't expand controls in the best match search, the one I want is visible thank you.

Indeed the whole search seems a tad inefficient as find_best_control_matches calls build_unique_dict on all the controls in a window (hundreds) to build a name map for finding. But seriously it can break out and bail the minute it finds a perfect match I would guess.

I'm hoping I can find a better way to get the control for this pane that is reliable and avoid using this best_match approach.

bernd-wechner:
Ironically, looking deeper:

`WindowSpecification.print_control_identifiers(self, depth=None, filename=None)`

accepts a depth. And yet, it applies a depth check only to the print stage, building a list of all controls to the lowest depth first. Hmmmm ... definite room for improvement here, to respect depth on the control walk, so that I might conceivably use depth as a means of finding Pane10 without ever seeing CombBoxes expand because they are on tier deeper than the the pane I'm after.

Ideally I suspect that the best_match criteria should include one called "expand = True" which we could pass to a best_match search or request in some manner, which conducts the best match search without expanding controls. Just on basis of the unexpanded controls.

But seriously it would be nice if the walk of controls also respected a depth setting rather than just the printing of them which does.

vasily-v-ryabov:
Hi @bernd-wechner thanks for all the suggestions. We aware about some methods which are too invasive. They are blockers for our script recorder architecture (PR #701) as well.

Reducing `best_match` algorithm complexity is another direction we're trying to work on. But it moves not so fast.

[...]

PR #733, opened 2019-05-15
issue #706: enumerate container items with FindItemByProperty.

airelil:
Use FindItemByProperty to access virtualized items
in ComboBox and ListView

### DIFF
--- a/pywinauto/controls/uia_controls.py
+++ b/pywinauto/controls/uia_controls.py
@@ -200,7 +200,11 @@ class ComboBoxWrapper(uiawrapper.UIAWrapper):
     # -----------------------------------------------------------
     def texts(self):
         """Return the text of the items in the combobox"""
-        texts = []
+        texts = self._texts_from_item_container()
+        if len(texts):
+            # flatten the list
+            return [ t for lst in texts for t in lst ]
+
         # ComboBox has to be expanded to populate a list of its children items
         try:
             super(ComboBoxWrapper, self).expand()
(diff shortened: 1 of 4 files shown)


pywinauto/controls/uia_controls.py
```diff
@@ -851,10 +855,19 @@ def get_item(self, row):
                     raise ValueError("Element '{0}' not found".format(row))
         elif isinstance(row, six.integer_types):
             # Get the item by a row index
-            # TODO: Can't get virtualized items that way
-            # TODO: See TODO section of item_count() method for details
-            list_items = self.children(content_only=True)
-            itm = list_items[self.__resolve_row_index(row)]
+            try:
+                com_elem = 0
```
vasily-v-ryabov: Is it possible that zero "handle" is passed into `UIAElementInfo.__init__`? Default root element assumes `None` otherwise it takes `ElementFromHandle(handle_or_elem)`. What `ElementFromHandle(0)` should return then? Will it raise `ComError`?

I'd suggest to use `None` and add check for `None` (raise ValueError in case of `None`).

airelil: Good point @vasily-v-ryabov, but it shouldn't be a case here, because we always call for ```get_elem_interface(com_elem, "VirtualizedItem").Realize()``` that already handles ValueError. But I added the case to the unit tests, so we should catch the problem if code is changed.

vasily-v-ryabov: OKI, it seems `com_elem = 0` is always overridden by at least one iteration of the loop below since you have `non_negative_number + 1` at the end of range. But in some hypothetical situation `self.__resolve_row_index(row)` may return `-1` and `com_elem = 0` will be passed to `get_elem_interface` and it will raise `AttributeError`. Well, it looks properly handled now.

pywinauto/uia_defines.py
```diff
@@ -218,6 +218,8 @@ class NoPatternInterfaceError(Exception):
 scroll_large_increment = IUIA().ui_automation_client.ScrollAmount_LargeIncrement
 scroll_small_increment = IUIA().ui_automation_client.ScrollAmount_SmallIncrement
 
+vt_empty = IUIA().ui_automation_client.VARIANT.empty.vt
```
cetygamer: These values are defined in win32defines.py.

airelil: Yep, you are right. But I intentionally didn't want to expose win32defines to any UIA modules. Probably too late, since there are already win32controls and common_controls imported in uia_controls... Honestly, I would move all these win32 hacks that have been added lately in UIA into a separate backend. Something like "hybrid".

vasily-v-ryabov: I think having third backend on Windows is too much. UIA can be optional, but Win32API is used for `Application` object and all other basic things on Windows. So I wouldn't support the idea to create a "hybrid" backend. Making UIA optional / separable (mostly for WinXP) should be enough.

This particular definition is OK to me in both variations. We can leave it as is since it's already merged.

Issue #735, opened 2019-05-19
Regression in unit tests of atspi branch for py2.7 failing on AppVeyor


airelil:
Half of py27 unit tests on atspi branch failing on AppVeyor:
https://ci.appveyor.com/project/pywinauto/pywinauto/builds/24639498/job/t3wa6ch92emgsa1t
I tried several failing tests to run locally, but they all passed. 
From Appveyor logs looks like ```application.start()``` got flaky.

Traceback examples:
```
Traceback (most recent call last):
  File "c:\python27-x64\lib\unittest\case.py", line 320, in run
    self.setUp()
  File "C:\projects\pywinauto\pywinauto\unittests\test_common_controls.py", line 102, in setUp
    self.ctrl = app.RowListSampleApplication.ListView.wrapper_object()
  File "C:\projects\pywinauto\pywinauto\base_application.py", line 191, in wrapper_object
    ctrls = self.__resolve_control(self.criteria)
  File "C:\projects\pywinauto\pywinauto\base_application.py", line 185, in __resolve_control
    raise e.original_exception
MatchError: Could not find 'RowListSampleApplication' in '['ToolTips', u'List', u'ListToolTips', u'Afx:00B70000:b:00010003:00000006:00040C6D']'
```
```
Traceback (most recent call last):
  File "c:\python27-x64\lib\unittest\case.py", line 320, in run
    self.setUp()
  File "C:\projects\pywinauto\pywinauto\unittests\test_common_controls.py", line 587, in setUp
    self.ctrl = self.app.MicrosoftControlSpy.TreeView.wrapper_object()
  File "C:\projects\pywinauto\pywinauto\base_application.py", line 191, in wrapper_object
    ctrls = self.__resolve_control(self.criteria)
  File "C:\projects\pywinauto\pywinauto\base_application.py", line 185, in __resolve_control
    raise e.original_exception
MatchError: Could not find 'MicrosoftControlSpy' in '[u'Dialog']'
```

airelil:
A short correction. I am able to reproduce failure on a local unittest. I made a mistake before and didn't run it on appropriate Python version.

airelil:
Should be fixed with PR #742

PR #740, opened 2019-05-25
Issue #735: fix py27-x64 unit tests for win32

airelil:
A quick fix to get win32 tests going. A better fix would be just align with the master.


airelil:
Better squash on merge, to remove temp debug commits.

airelil:
Closing since merging PR #742

Issue #741, opened 2019-05-25
Add support for AtspiDocument interface


airelil:
Creating this issue to define requirements for AtsiDocument support as below:
- Add or extend an existing GTK sample app with controls supporting AtspiDocument interface.
- Add low-level interface class AtspiDocument in atspi_objects.py
- Add support of AtspiDocument interface in atspi_element_element_info.py

PR #771, opened 2019-07-12
Issue #741: add support for AtspiDocument interface

airelil:
Access [AtspiDocument](https://developer.gnome.org/libatspi/stable/libatspi-atspi-document.html#AtspiDocument) Interface as ```AtspiElementInfo.document``` property.

I couldn't create a simple pyGTK example with AtspiDocument interface. Probably, it has to contain a UI control with embedded browser or PDF viewer. So the unit tests use mocked low-level GLib functions instead. Also, interfacing to [g_hash_table](https://developer.gnome.org/glib/stable/glib-Hash-Tables.html)  returned by ```atspi_document_get_attributes()``` is simplified and assumes table's key/value types as strings.


pywinauto/linux/atspi_objects.py
```diff
@@ -491,7 +502,67 @@ def get_iface_func(self, func_name):
...
+_g_hash_table_destroy = _glib.g_hash_table_destroy
+_g_hash_table_destroy.restype = None
+_g_hash_table_destroy.argtypes = [c_void_p]
+
+_g_hash_table_insert = _glib.g_hash_table_insert
+_g_hash_table_insert.restype = c_bool
+_g_hash_table_insert.argtypes = [c_void_p, c_void_p, c_void_p]
+
+
+def _ghash2dic(ghash):
```
vasily-v-ryabov: Maybe this function could become operator `__getitem__` of class `GHashTable`.

pywinauto/linux/atspi_objects.py
```diff
@@ -986,3 +1079,53 @@ def get_minimum_increment(self, g_error_pointer=None):
...
+    _get_attributes.restype = c_void_p
+
+    def __init__(self, pointer):
+        """Init the ATSPI Document Interface"""
+        self._pointer = pointer
+
+    @g_error_handler
+    def get_locale(self, g_error_pointer=None):
+        """
+        Gets the locale associated with the document's content, e.g. the locale for LOCALE_TYPE_MESSAGES.
```
vasily-v-ryabov: (minor) All docstrings in pywinauto use infinitive form of a verb (without "s").

pywinauto/linux/atspi_objects.py
```diff
@@ -491,7 +502,67 @@ def get_iface_func(self, func_name):
             return None
 
 
+class GLIB(IATSPI):
+
+    """Python wrapper around C functions from GLib library"""
+
+    LIB = "libglib-2.0.so"
```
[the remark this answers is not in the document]
airelil: Actually, am thinking about setting up a stand-alone ```_find_library()``` utility function that will get the list of library names and try to load the first available. Re-use the helper for loading atspi and GLib shared libs.

pywinauto/linux/atspi_objects.py
```diff
@@ -491,7 +502,67 @@ def get_iface_func(self, func_name):
...
+_g_hash_table_foreach.restype = None
+_g_hash_table_foreach.argtypes = [c_void_p, _GHFunc, c_void_p]
+
+_g_hash_table_destroy = _glib.g_hash_table_destroy
+_g_hash_table_destroy.restype = None
+_g_hash_table_destroy.argtypes = [c_void_p]
+
+_g_hash_table_insert = _glib.g_hash_table_insert
+_g_hash_table_insert.restype = c_bool
+_g_hash_table_insert.argtypes = [c_void_p, c_void_p, c_void_p]
```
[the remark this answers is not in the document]
airelil: I was following the same pattern we have in ```win32functions```, just a set of interfaces to native functions, but it might look good if grouped logically under a class. However, not sure about routing everything through a method similar to ```get_iface_func``` as there is no ambibguity in Glib function names like in ATSPI interfaces.

pywinauto/linux/atspi_objects.py
airelil: Noted

pywinauto/linux/atspi_objects.py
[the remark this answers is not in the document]
airelil: BTW, hope you noticed, it's strictly specialized for strings and quite far from a full fledged python wrapper arount g_hash_table.

pywinauto/linux/atspi_objects.py
airelil: Hm, it's only a converter that returns the whole dictionary at once...

vasily-v-ryabov:
Let's remember the discussion points. Merging for now.

Issue #783, opened 2019-07-26
Implement AtspiImage


airelil:
For the Linux platform we need to suppor AtspiImage as descibed: https://developer.gnome.org/libatspi/stable/libatspi-atspi-image.html#AtspiImage

PR #794, opened 2019-08-17
Issue #783

airelil:
- Add support for `AtspiImage` interface
- Add `AtspiElementInfor.image` property
- Add `atspi_controls.ImageWrapper`
- Fixes and refactoring in in atspi_objects
- Move common code fore POINT and Structure into `base_types` (RECT is next)
- Fix TreeViewAdditionalTestCases.testCheckBoxes
- AtspiImages unit tests


pywinauto/controls/atspi_controls.py
```diff
@@ -278,3 +278,37 @@ def select(self, start=0, end=None):
...
+        """Get image description"""
+        self.element_info.image_get_description()
+
+    def get_locale(self):
+        """Get image locale"""
+        self.element_info.image_get_locale()
+
+    def get_size(self):
+        """Get image size.
+        Return a tupe with width and height"""
```
vasily-v-ryabov: Typo: `type` -> `tuple`. Also empty line after sumary is necessary with wrapping closing quotes to next line.

pywinauto/controls/atspi_controls.py
```diff
@@ -278,3 +278,37 @@ def select(self, start=0, end=None):
...
+        pnt = self.element_info.image_get_size()
+        return (pnt.x, pnt.y)
+
+    def get_bounding_box(self):
+        """Get image bounding box"""
+        return self.element_info.image_get_extents()
+
+    def get_position(self):
```
vasily-v-ryabov: Maybe `get_` is redundant in all method names. Not sure it's consistent with method names for other wrappers, but we have method `rectangle` instead of `get_rectangle` for example.

pywinauto/linux/atspi_element_info.py
```diff
@@ -239,3 +245,32 @@ def document_get_attribute_value(self, attrib):
...
+
+    def image_get_position(self):
+        """Return the image position"""
+        return self.image.get_position()
+
+    def image_get_extents(self):
+        """Return the image extents"""
+        return self.image.get_extents()
+
+    def image_get_size(self):
```
vasily-v-ryabov: Do we really need all these image related properties for search class `AtspiElementInfo`? In the future we have plans to use `__slots__` static member for `ElementInfo` derived classes so that memory footprint is minimal.

airelil: Good point. Will move it directly into atspi_controls.ImageWrapper. Probably even `image` property itself.

pywinauto/controls/atspi_controls.py
airelil: Resolved with e358ab2

pywinauto/controls/atspi_controls.py
airelil: Resolved with e358ab2

pywinauto/linux/atspi_element_info.py
vasily-v-ryabov: If we leave `image` property in `AtspiElementInfo` we should think how it looks for end user. I mean something like this: `.child_window(image=AtspiImage(some_pointer))`. Maybe such low level pointer is not a good thing to show publicly. But some text property from image object could be useful for search.

For example `.child_window(image_descr="some text")` might be really useful.

vasily-v-ryabov: The same consideration makes sense for AtspiDocument interface as well. I'd suggest to file a separate issue on this where we can discuss `AtspiElementInfo` interface minimization.

airelil: Agree. Created issue #806 to continue the discussion.

Issue #802, opened 2019-08-31
[ValueError: string too long] Strings should be unicode-escaped before calculating its size in set_edit_text


SaadeMotion:
When using emojis in a text, eg.
```
Hey You! ⛔
You`re dying ⚰
Buy your casket now! at this reseller 📍
```

You should calculate string size after unicode-escaping it in:
https://github.com/pywinauto/pywinauto/blob/2d11804cb36b3cee70af26ab6a05e4a299e80250/pywinauto/controls/win32_controls.py#L819

Otherwise ctypes will throw a ` ValueError: string too long`.

I just bypassed this by:
```python
def set_caption(text):
    buffer = ctypes.create_unicode_buffer(text, size=len(text.encode('unicode_escape'))) # Here's the fix
    app.window(auto_id="tbCaption").send_message( win32con.WM_SETTEXT, True, ctypes.byref(buffer) )
```

instead of:
```python
def set_caption(text):
    app.window(auto_id="tbCaption").set_text( text )
```

PR #803, opened 2019-08-31
Issue #802

saade:
As described in #802 , here's the fix for set_edit_text

### DIFF
--- a/pywinauto/controls/win32_controls.py
+++ b/pywinauto/controls/win32_controls.py
@@ -816,9 +816,9 @@ class EditWrapper(hwndwrapper.HwndWrapper):
                 aligned_text = six.binary_type(text)
 
         if isinstance(aligned_text, six.text_type):
-            buffer = ctypes.create_unicode_buffer(aligned_text, size=len(aligned_text) + 1)
+            buffer = ctypes.create_unicode_buffer(aligned_text, size=len(aligned_text.encode('unicode_escape')) + 1)
         else:
-            buffer = ctypes.create_string_buffer(aligned_text, size=len(aligned_text) + 1)
+            buffer = ctypes.create_string_buffer(aligned_text, size=len(aligned_text.encode('unicode_escape')) + 1)
 
         # replace the selection with
         self.send_message(win32defines.EM_REPLACESEL, True, ctypes.byref(buffer))


pywinauto/controls/win32_controls.py
```diff
@@ -816,9 +816,9 @@ def set_edit_text(self, text, pos_start = None, pos_end = None):
                 aligned_text = six.binary_type(text)
 
         if isinstance(aligned_text, six.text_type):
-            buffer = ctypes.create_unicode_buffer(aligned_text, size=len(aligned_text) + 1)
+            buffer = ctypes.create_unicode_buffer(aligned_text, size=len(aligned_text.encode('unicode_escape')) + 1)
         else:
-            buffer = ctypes.create_string_buffer(aligned_text, size=len(aligned_text) + 1)
+            buffer = ctypes.create_string_buffer(aligned_text, size=len(aligned_text.encode('unicode_escape')) + 1)
```
airelil: @SaadeMotion , you might noticed a failure in pywinauto.unittests.test_hwndwrapper.NotepadRegressionTests.testMenuSelectNotepad_bug 
```
Error message:
'ascii' codec can't decode byte 0xef in position 33: ordinal not in range(128)
Error stack trace:
  File "c:\python27\lib\unittest\case.py", line 329, in run
    testMethod()
  File "C:\projects\pywinauto\pywinauto\unittests\test_hwndwrapper.py", line 837, in testMenuSelectNotepad_bug
    self.app2.UntitledNotepad.Edit.set_edit_text(text)
  File "C:\projects\pywinauto\pywinauto\controls\win32_controls.py", line 821, in set_edit_text
    buffer = ctypes.create_string_buffer(aligned_text, size=len(aligned_text.encode('unicode_escape')) + 1)
UnicodeDecodeError: 'ascii' codec can't decode byte 0xef in position 33: ordinal not in range(128)
```
It seems that the fix should be applied only for a unicode buffer, but not for simple strings

airelil:
Also, as per [codecov report](https://codecov.io/gh/pywinauto/pywinauto/compare/2d11804cb36b3cee70af26ab6a05e4a299e80250...f0fce8805507adbedc13f9edeb15a7c91bcca17c/changes) for win32_controls.py the `if` branch with `create_unicode_buffer` is not covered by our tests. We'd better add a simple test for it.

Issue #810, opened 2019-09-05
Potential typo bug in pywinauto top-level file.


pakal:
Line 91 in `pywinauto/__init__.py` we have this code which looks like a typo ("ElementNotFoundError = findwindows.ElementNotFoundError" is duplicated as is), wasn't it supposed to be WindowNotFoundError instead ?

```
    WindowAmbiguousError = findwindows.WindowAmbiguousError
    ElementNotFoundError = findwindows.ElementNotFoundError

    if UIA_support:
        ElementNotFoundError = findwindows.ElementNotFoundError
        ElementAmbiguousError = findwindows.ElementAmbiguousError
```

vasily-v-ryabov:
It seems you're right for `ElementNotFoundError` which was `WindowNotFoundError` a long time ago (when there was no backend concept and Win32 API was the only way).

`WindowAmbiguousError` is not a duplication, but an old alias for backward compatibility. Anyway all this could be cleaned up in current re-factoring. Easy to fix as your first issue. ;)

pakal:
Allright, I'm having a quick look at it -> https://github.com/pywinauto/pywinauto/pull/811

PR #811, opened 2019-09-06
Fix missing top-level import "WindowNotFoundError".

pakal:
Fixes https://github.com/pywinauto/pywinauto/issues/810

I didn't find a proper place for a quick non-regression test so I added a test file.

### DIFF
--- a/pywinauto/__init__.py
+++ b/pywinauto/__init__.py
@@ -89,7 +89,7 @@ if sys.platform == 'win32':
     from . import findwindows
 
     WindowAmbiguousError = findwindows.WindowAmbiguousError
-    ElementNotFoundError = findwindows.ElementNotFoundError
+    WindowNotFoundError = findwindows.WindowNotFoundError
 
     if UIA_support:
         ElementNotFoundError = findwindows.ElementNotFoundError
--- a/pywinauto/unittests/test_architecture.py
+++ b/pywinauto/unittests/test_architecture.py
@@ -1,0 +1,20 @@
+import unittest
+
+from pywinauto import UIA_support
+
+
+class PublicImportsTests(unittest.TestCase):
+
+    def test_top_level_imports(self):
+        if UIA_support:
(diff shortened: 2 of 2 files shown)


pywinauto/__init__.py
```diff
@@ -89,7 +89,7 @@ def _get_com_threading_mode(module_sys):
     from . import findwindows
 
     WindowAmbiguousError = findwindows.WindowAmbiguousError
-    ElementNotFoundError = findwindows.ElementNotFoundError
+    WindowNotFoundError = findwindows.WindowNotFoundError
 
     if UIA_support:
```
airelil: Honestly, not sure if  we still need this condition. Might worth removing.

vasily-v-ryabov:
Also the commit seems done with another e-mail not matched with your GitHub account. It can be resolved by "amend last commit" + "force push" (the first way) or by adding this e-mail to your GitHub account as a secondary one. Both ways allow counting your contribution properly.

pywinauto/__init__.py
pakal: Hm, I guess such aliases are completely harmless, and removing them might break third-party packages which use find_window() and catch this exception... I'd highly suggest never breaking compatibility in minor versions, as per semantic versioning, since we have no idea of the impact on end users.

pakal:
Alright I added my (current client's) email to my github account, I'm updating this PR asap.

pywinauto/__init__.py
airelil: I meant `if UIA_support` condition. We might just set ElementNotFoundError, ElementAmbiguousError uncoditionally.

pakal: Whups, indeed now that you mention it, I don't find the relationship between UIA_support and these exceptions in nowadays's code... but the PR is already merged.

pakal:
I opened https://github.com/pywinauto/pywinauto/pull/816 just in case, C.I will decide if UIA_support condition was indeed not necessary anymore here.

PR #815, opened 2019-09-09
Add use_magic_lookup boolean attribute to WindowSpecification and others (WIP)

pakal:
This would allow disabling magic getattr() lookup of child widgets, which in some applications (with lots of similar elements) is more confusing than handy.

### DIFF
--- a/pywinauto/application.py
+++ b/pywinauto/application.py
@@ -141,11 +141,12 @@ class WindowSpecification(object):
                          'active': ('is_active',),
                          }
 
-    def __init__(self, search_criteria):
+    def __init__(self, search_criteria, use_magic_lookup=True):
         """
         Initialize the class
 
         :param search_criteria: the criteria to match a dialog
+        :param use_magic_lookup: whether attribute access must turn into child_window(best_match=...) search as fallback
         """
         # kwargs will contain however to find this window
         if 'backend' not in search_criteria:
@@ -158,6 +159,7 @@ class WindowSpecification(object):
         self.criteria = [search_criteria, ]
         self.actions = ActionLogger()
         self.backend = registry.backends[search_criteria['backend']]
+        self.use_magic_lookup = use_magic_lookup
 
         if self.backend.name == 'win32':
             # Non PEP-8 aliases for partial backward compatibility
@@ -173,7 +175,7 @@ class WindowSpecification(object):
             self.window_ = deprecated(self.child_window, deprecated_name='window_')
 
     def __call__(self, *args, **kwargs):
-        """No __call__ so return a usefull error"""
+        """No __call__ so return a useful error"""
         if "best_match" in self.criteria[-1]:
             raise AttributeError("Neither GUI element (wrapper) " \
                 "nor wrapper method '{0}' were found (typo?)".
@@ -182,8 +184,8 @@ class WindowSpecification(object):
         message = (
             "You tried to execute a function call on a WindowSpecification "
             "instance. You probably have a typo for one of the methods of "
-            "this class.\n"
-            "The criteria leading up to this is: " + str(self.criteria))
+            "this class or of the targeted wrapper object.\n"
+            "The criteria leading up to this are: " + str(self.criteria))
 
         raise AttributeError(message)
 
@@ -277,7 +279,7 @@ class WindowSpecification(object):
         if 'top_level_only' not in criteria:
             criteria['top_level_only'] = False
 
-        new_item = WindowSpecification(self.criteria[0])
+        new_item = WindowSpecification(self.criteria[0], use_magic_lookup=self.use_magic_lookup)
         new_item.criteria.extend(self.criteria[1:])
         new_item.criteria.append(criteria)
 
@@ -307,7 +309,7 @@ class WindowSpecification(object):
         """
         # if we already have 2 levels of criteria (dlg, control)
         # then resolve the control and do a getitem on it for the
-        if len(self.criteria) >= 2:
+        if len(self.criteria) >= 2:  # FIXME - this is surprising
 
             ctrls = self.__resolve_control(self.criteria)
 
@@ -324,7 +326,7 @@ class WindowSpecification(object):
 
         # if we get here then we must have only had one criteria so far
         # so create a new :class:`WindowSpecification` for this control
-        new_item = WindowSpecification(self.criteria[0])
+        new_item = WindowSpecification(self.criteria[0], use_magic_lookup=self.use_magic_lookup)
 
         # add our new criteria
         new_item.criteria.append({"best_match": key})
@@ -345,6 +347,21 @@ class WindowSpecification(object):
         Otherwise delegate functionality to :func:`__getitem__` - which
         sets the appropriate criteria for the control.
         """
+        use_magic_lookup = object.__getattribute__(self, "use_magic_lookup")  # Beware of recursions here!
+        if not use_magic_lookup:
+            try:
+                return object.__getattribute__(self, attr_name)
+            except AttributeError:
+                wrapper_object = self.wrapper_object()
+                try:
+                    return getattr(wrapper_object, attr_name)
+                except AttributeError:
+                    message = (
+                        'Attribute "%s" exists neither on %s object nor on'
+                        'targeted %s element wrapper (typo? or set use_magic_lookup to True?)' %
+                        (attr_name, self.__class__, wrapper_object.__class__))
+                    raise AttributeError(message)
+
         if attr_name in ['__dict__', '__members__', '__methods__', '__class__', '__name__']:
             return object.__getattribute__(self, attr_name)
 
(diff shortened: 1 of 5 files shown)


vasily-v-ryabov:
Hmm... The implementation looks good technically. However I would think about `.dump_tree()` output in case of disabled magic lookup. It will still print magic names and will spend resources on getting them.

Also I would think about consistency between keywords `magic_lookup` and `best_match`. It might be not obvious that it means the same thing. Maybe `use_best_match` is more consistent if we decide not touching legacy keyword.

pywinauto/unittests/test_application.py
```diff
@@ -46,6 +46,9 @@
 import ctypes
 import mock
 
+from pywinauto.controls.common_controls import TrackbarWrapper
+from pywinauto.controls.uiawrapper import UIAWrapper
```
airelil: Pls remember that 'UIA' is still optional we use it directly only in the dedicated files like test_uiawrapper.py  Try to see if your unit tests can be split into win32 and uia parts. Use `if UIA_support:` to verify. See the failing unit test on AppVeyor:
```
  File "c:\python27\lib\unittest\case.py", line 329, in run
    testMethod()
  File "c:\python27\lib\site-packages\nose\loader.py", line 418, in loadTestsFromName
    addr.filename, addr.module)
  File "c:\python27\lib\site-packages\nose\importer.py", line 47, in importFromPath
    return self.importFromDir(dir_path, fqname)
  File "c:\python27\lib\site-packages\nose\importer.py", line 94, in importFromDir
    mod = load_module(part_fqname, fh, filename, desc)
  File "C:\projects\pywinauto\pywinauto\unittests\test_application.py", line 50, in <module>
    from pywinauto.controls.uiawrapper import UIAWrapper
  File "C:\projects\pywinauto\pywinauto\controls\uiawrapper.py", line 40, in <module>
    import comtypes
ImportError: No module named comtypes
```

pywinauto/unittests/test_application.py
```diff
@@ -46,6 +46,9 @@
 import ctypes
 import mock
 
+from pywinauto.controls.common_controls import TrackbarWrapper
```
airelil: Both new imports should be moved below `sys.path.append('.')` 
Helps to run individual tests from command line:
```
python pywinauto\unittests\test_application.py DesktopWin32WindowSpecificationTests.test_non_magic_desktop
```

pywinauto/unittests/test_application.py
pakal: Thanks I'll have a look at it, Pycharm autoimports got me ^^

pakal:
Regarding dump_tree(): It depends, I'm not removing the best_match system entirely, just the magic lookup via __getattribute__(). 
So it can still be useful to do a .child_window(best_match=xxx) with the help of dump_tree(), it's just that this call is explicit and not auto-triggered by attribute access on WindowSpecification objects and others.
Yet, use_magic_lookup is maybe not the most obvious name for this parameters, let's think about alternatives.

pywinauto/unittests/test_application.py
```diff
@@ -1358,6 +1419,25 @@ def test_top_from_point_win32(self):
...
+        self.assertEqual(self.desktop_no_magic.use_magic_lookup, False)
+
+        window = self.desktop_no_magic.window(title=self.window_title, process=self.app.process)
+        self.assertEqual(window.use_magic_lookup, False)
+
+        dlg = window.child_window(class_name="msctls_trackbar32").wrapper_object()
+        self.assertIsInstance(dlg, TrackbarWrapper)
+
+        pos = dlg.get_position()
+        self.assertIsInstance(pos, int)
```
airelil: Probably already noticed. This fails on Python 2.7 x64
```
  File "c:\python27-x64\lib\unittest\case.py", line 329, in run
    testMethod()
  File "C:\projects\pywinauto\pywinauto\unittests\test_application.py", line 1433, in test_non_magic_desktop
    self.assertIsInstance(pos, int)
  File "c:\python27-x64\lib\unittest\case.py", line 968, in assertIsInstance
    self.fail(self._formatMessage(msg, standardMsg))
  File "c:\python27-x64\lib\unittest\case.py", line 410, in fail
    raise self.failureException(msg)
AssertionError: 0L is not an instance of <type 'int'>
```

airelil: Might reuse `six`:
```
        import six
...
...
        self.assertIsInstance(pos, six.integer_types)

```

vasily-v-ryabov:
Well, you're right about `best_match` and `.dump_tree()`. Maybe `use_attr_magic_lookup` is a little bit more clear then.

pakal:
Alright "use_magic_attr_lookup" then ?

I'm having weird crashes on test teardown, which is not certainly related to my changes, like if some windows had "None" instead of their title, crashing the difflib operations:

```
Error
  File "C:\Python27\lib\unittest\case.py", line 358, in run
    self.tearDown()
  File "workspace\pywinauto\pywinauto\unittests\test_application.py", line 1293, in tearDown
    self.desktop.MFC_samplesDialog.wait_not('exists')
  File "workspace\pywinauto\pywinauto\application.py", line 567, in wait_not
    lambda: not self.__check_all_conditions(check_method_names, retry_interval))
  File "workspace\pywinauto\pywinauto\timings.py", line 359, in wait_until
    func_val = func(*args, **kwargs)
  File "workspace\pywinauto\pywinauto\application.py", line 567, in <lambda>
    lambda: not self.__check_all_conditions(check_method_names, retry_interval))
  File "workspace\pywinauto\pywinauto\application.py", line 476, in __check_all_conditions
    if not check(retry_interval, float(retry_interval) // 2):
  File "workspace\pywinauto\pywinauto\application.py", line 427, in exists
    self.__resolve_control(exists_criteria, timeout, retry_interval)
  File "workspace\pywinauto\pywinauto\application.py", line 258, in __resolve_control
    criteria)
  File "workspace\pywinauto\pywinauto\timings.py", line 436, in wait_until_passes
    func_val = func(*args, **kwargs)
  File "workspace\pywinauto\pywinauto\application.py", line 203, in __get_ctrl
    dialog = self.backend.generic_wrapper_class(findwindows.find_element(**criteria[0]))
  File "workspace\pywinauto\pywinauto\findwindows.py", line 84, in find_element
    elements = find_elements(**kwargs)
  File "workspace\pywinauto\pywinauto\findwindows.py", line 303, in find_elements
    elements = findbestmatch.find_best_control_matches(best_match, wrapped_elems)
  File "workspace\pywinauto\pywinauto\findbestmatch.py", line 507, in find_best_control_matches
    best_ratio, best_texts = name_control_map.find_best_matches(search_text)
  File "workspace\pywinauto\pywinauto\findbestmatch.py", line 419, in find_best_matches
    ratio_calc.set_seq2(text)
  File "C:\Python27\lib\difflib.py", line 285, in set_seq2
    self.__chain_b()
  File "C:\Python27\lib\difflib.py", line 318, in __chain_b
    for i, elt in enumerate(b):
TypeError: 'NoneType' object is not iterable
```

The name_control_map indeed contains an entry like this :
u'': <uiawrapper.UIAWrapper - '', None, 0>,

I can't investigate more just now, but in case someone has an idea...

I'll need to add docs for my changes too.

vasily-v-ryabov:
OK, I would suggest `allow_magic_lookup`. `attr` doesn't cover item access case.

The last error usually happens when some `.window_text()` or `.element_info.name` property returned `None` while "best_match" algorithm expects string objects only.

pakal:
In the current PR, I've left "_getitem_" as is, since it doesn't interfere with any "default" behaviour of objects; having item lookup as a shortcut to child_window(best_match=xxx) looks handy and harmless, unless some objects of the chain (element_info, or its subelements...) already have item access (but it doesn't seem so). So allow_magic_lookup and allow_magic_attr_lookup would both be True, the latter being more accurate, and the former opening more opportunities for further refuinements. I have to think about it a bit more.

Regarding the error with None, is it an error to be fixed in .window_text() or .element_info.name, or should we pragmatically just filter non-string values out of the name mapper to ensure these crashs stop?

vasily-v-ryabov:
`allow_magic_lookup` looks more compact. I'd vote for it finally.

> Regarding the error with None, is it an error to be fixed in .window_text() or .element_info.name, or should we pragmatically just filter non-string values out of the name mapper to ensure these crashes stop?

Until now we tried to always return empty string if something goes wrong. But filtering `None` values out would be better solution probably.

airelil:
```
  File "workspace\pywinauto\pywinauto\unittests\test_application.py", line 1293, in tearDown
    self.desktop.MFC_samplesDialog.wait_not('exists')
```
I think using `wait_not(`exists`)` is the key here. Why didn't you stick to `wait_not('visible`)` ?

vasily-v-ryabov:
@airelil it shouldn't matter because `None` is returned as text for some other window in the system. We don't see this case on AppVeyor, because there is no side apps running in the background there.

pakal:
OK I've pushed a bugfix for "None" window names, and some documentation for the new "allow_magic_lookup" parameter.

@airelil - I switched from "visible" to "exists" because for me "visible" failed when the dialog didn't exist at all, which was the case because it closed very quickly after the test. Let's see if the CI complains because of this.

pywinauto/application.py
```diff
@@ -307,7 +309,7 @@ def __getitem__(self, key):
         """
         # if we already have 2 levels of criteria (dlg, control)
         # then resolve the control and do a getitem on it for the
-        if len(self.criteria) >= 2:
+        if len(self.criteria) >= 2:  # FIXME - this is surprising
```
vasily-v-ryabov: This constraint should be for `backend="win32"` only. We missed additional check of backend name here. Thanks for catching it! We may discuss it in separate issue though.

vasily-v-ryabov: Oh, nevermind, this is something different. Anyway let's discuss it later.

Issue #798, opened 2019-08-27
Improper log() arguments in codebase


pakal:
Hello,

Several calls to `ActionLogger().log()` follow the "print" syntax (comma-separated elements to concatenate) and not the logging syntax (msg with placeholder and then its formatting arguments).

For example in remote_memory_block.py:

```            
ActionLogger().log('Error: Write failed: address = ', address)
last_error = win32api.GetLastError()
ActionLogger().log('Error: LastError = ', last_error, ': ',
         win32api.FormatMessage(last_error).rstrip())
```   

This code seems to be rarely hit, but it will raise a "TypeError: not all arguments converted during string formatting" if it is encountered. Otehr log() calls use str.format() to directly issue a single string.

PS: thanks for the awesome GUI test framework

airelil:
Thanks for the report, It seems to be remnants of using the `_CustomLogger`. We, definitely, should have a look into it

TomRobo237:
From the looks of it, here's where all the `ActionLogger()` calls are:
```
pywinauto $ for i in $(grep -r . -e'ActionLogger().log' | awk 'BEGIN {FS=":"};{print$1}' | sort -u) ; do echo $i ; nl $i | awk '/ActionLogger\(\)\.log/ {print $0}' ; done
./pywinauto/controls/hwndwrapper.py
  1756      ActionLogger().log(message)
./pywinauto/findbestmatch.py
   317              pass #ActionLogger().log('Warning! Cannot get control.texts()') #\nTraceback:\n' + traceback.format_exc())
./pywinauto/handleprops.py
    81          ActionLogger().log('WARNING! Cannot retrieve text length for handle = ' + str(handle))
./pywinauto/remote_memory_block.py
   127              ActionLogger().log('================== Error: Failed to write guard signature: address = ' +
   130              ActionLogger().log('LastError = ' + str(last_error) + ': ' + win32api.FormatMessage(last_error).rstrip())
   138              ActionLogger().log('Warning: cannot close process handle!')
   164              pass  # ActionLogger().log('\nWARNING: Cannot call VirtualFreeEx! process_id == 0.')
   208              ActionLogger().log('Error: Write failed: address = ', address)
   210              ActionLogger().log('Error: LastError = ', last_error, ': ',
   259                      ActionLogger().log('Read: WARNING! self.mem_address =' +
   261                      ActionLogger().log('LastError = ' + str(last_error) +
[...]

PR #829, opened 2019-09-26
Converted ActionLogger().log commands from print style to string concat

TomRobo237:
Fixed Action logging in print style by swapping to strings as described in issue #798

### DIFF
--- a/pywinauto/remote_memory_block.py
+++ b/pywinauto/remote_memory_block.py
@@ -205,9 +205,9 @@ class RemoteMemoryBlock(object):
         )
 
         if ret == 0:
-            ActionLogger().log('Error: Write failed: address = ', address)
+            ActionLogger().log('Error: Write failed: address = ' + str(address))
             last_error = win32api.GetLastError()
-            ActionLogger().log('Error: LastError = ', last_error, ': ',
+            ActionLogger().log('Error: LastError = ' + str(last_error) + ': ' +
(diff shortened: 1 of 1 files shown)


airelil:
Thanks for the contribution! As I understand, we didn't bump into the problem before because this piece of code never was executed.  Ideally, l'd like to see a unit test for the case.

TomRobo237:
Hmmm, writing that may be a little beyond me at this point. I've only really begun to play around with some of the pywinauto features. I'm not sure what I would write to hit that code block tbh.

pakal:
Nice, the PR seems to solve the problem (although %-formatting or str.format() would be more readable, but I'm nitpicking).

Indeed managing to trigger an OS error on remote process WriteProcessMemory() is not easy, does anyone know how/when it can happen ?

airelil:
I, usually just use 'mock' to return the errors I need to get to hit the code. There are several examples for mock usage in unittests directory.

vasily-v-ryabov:
Thanks for the fix. It aligns all such statements across the module. So I decided to merge it without unit tests. It seems not a critical point for this rare case.

vasily-v-ryabov:
@pakal `WriteProcessMemory` usually fails when calling it in low privilege process trying to write into admin process. But recent pywinauto versions do all checks for running Python as administrator when connecting to admin-privileged process. So this particular case may be more interesting.

PR #831, opened 2019-10-06
Fix ctypes.windll usage conflicts with other libraries

vasily-v-ryabov:
* `windll = LibraryLoader(WinDLL)` isolates cached Win32 libraries for pywinauto usage only. Other library may pollute `ctypes.windll` without affecting any pywinauto code.
 * All Win32 function definitions have been moved from `keyboard.py` to `win32functions.py`.
 * Function names use explicit Unicode suffix `W` like `MapVirtualKeyW` instead of `MapVirtualKey`.
 * Add recent changes to 0.6.8 release notes.

### DIFF
--- a/docs/HISTORY.txt
+++ b/docs/HISTORY.txt
@@ -2,6 +2,26 @@
 Change Log
 ==========
 
+0.6.8 Bug Fixes
+--------------------------------------------------------------------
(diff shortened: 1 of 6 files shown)


airelil:
Can't merge on mobile, just approving

PR #833, opened 2019-10-07
handle case when window title is None

philmbailey:
Not sure if you'd prefer this change or to skip the encoding step below. Thoughts appreciated.  Thanks!

### DIFF
--- a/pywinauto/handleprops.py
+++ b/pywinauto/handleprops.py
@@ -62,7 +62,7 @@ def text(handle):
     if class_name == 'MSCTFIME UI':
         return 'M'
     if class_name is None:
-        return None
+        return ''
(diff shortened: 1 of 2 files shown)


vasily-v-ryabov:
Hi @philmbailey thanks for your contribution. One more thing to improve is that commits were done with e-mail not matched with your GitHub account. There are 2 possible ways to count your contribution properly:

1. Add alternative e-mail to your account so it will match with your profile.
2. Amend all commits with proper e-mail and force push the branch (or re-create the branch with new commits if it's simpler for you).

philmbailey:
Thanks! The email address should be philmb@gmail.com. I believe that is the one in the latest commit in the PR.

PR #834, opened 2019-10-11
add vk_packet option to send_keys

philmbailey:
This change is useful to me as I have an application that only handles WM_KEYDOWN for hotkeys and not WM_CHAR. I'd like to be able to automate key presses via pywinauto. Thanks for your consideration.

### DIFF
--- a/pywinauto/keyboard.py
+++ b/pywinauto/keyboard.py
@@ -98,10 +98,19 @@ Use curly brackers to escape modifiers and type reserved symbols as single keys:
     send_keys('{^}a{^}c{%}') # type string "^a^c%" (Ctrl will not be pressed)
     send_keys('{{}ENTER{}}') # type string "{ENTER}" without pressing Enter key
 
+For Windows only, pywinauto defaults to sending a virtual key packet
+(VK_PACKET) for textual input.  For applications that do not handle VK_PACKET
+appropriately, the ``vk_packet`` option may be set to ``False``.  In this case
+pywinauto will attempt to send the virtual key code of the requested key.  This
+option only affects the behavior of keys matching [-=[]\;',./a-zA-Z0-9 ].  Note
+that upper and lower case are included for a-z.  Both reference the same
+virtual key for convenience.
+
 """
 from __future__ import unicode_literals
 
 import sys
+import string
 
 from . import deprecated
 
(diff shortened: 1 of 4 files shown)


airelil:
@philmbailey, sorry for the delay, will be able to make a proper review only the next weekend. BTW, do you have in mind any other use cases except of your app?

philmbailey:
None that I know of specifically. This seemed generally useful to me in that it should more closely emulate the event stream generated by a physical key press.

pywinauto/keyboard.py
```diff
@@ -302,6 +311,29 @@
         '%': VK_MENU,
     }
 
+    # Virtual keys that map to an ASCII character
+    # See https://docs.microsoft.com/en-us/windows/win32/inputdev/virtual-key-codes
+    ASCII_VK = {
```
vasily-v-ryabov: I'd suggest to stick this name to PEP8: `ascii_vk`.

pywinauto/keyboard.py
```diff
@@ -302,6 +311,29 @@
...
 
+    # Virtual keys that map to an ASCII character
+    # See https://docs.microsoft.com/en-us/windows/win32/inputdev/virtual-key-codes
+    ASCII_VK = {
+        ' ': 0x20,
+        '=': 0xbb,
+        ',': 0xbc,
+        '-': 0xbd,
+        '.': 0xbe,
+        # According to the above reference, the following variy per region.
```
vasily-v-ryabov: Typo: `variy` -> `varies`.

vasily-v-ryabov:
Less risky option for me is having `vk_packet=False` by default to keep legacy behavior. We can switch to `vk_packet=True, with_spaces=True` default values in 0.7.0 since we're breaking much more compatibility cases in major release.

pywinauto/keyboard.py
```diff
@@ -98,10 +98,19 @@
     send_keys('{^}a{^}c{%}') # type string "^a^c%" (Ctrl will not be pressed)
     send_keys('{{}ENTER{}}') # type string "{ENTER}" without pressing Enter key
 
+For Windows only, pywinauto defaults to sending a virtual key packet
+(VK_PACKET) for textual input.  For applications that do not handle VK_PACKET
+appropriately, the ``vk_packet`` option may be set to ``False``.  In this case
+pywinauto will attempt to send the virtual key code of the requested key.  This
+option only affects the behavior of keys matching [-=[]\;',./a-zA-Z0-9 ].  Note
+that upper and lower case are included for a-z.  Both reference the same
+virtual key and are included for convenience.
```
vasily-v-ryabov: Looks like a typo: "Both reference the same virtual key and are included for convenience." -> I'd suggest "Both reference the same virtual key for convenience."

PR #844, opened 2019-10-31
Merge ATSPI branch with the latest master

airelil:


pywinauto/controls/atspi_controls.py
```diff
@@ -0,0 +1,345 @@
...
+
+    def click(self):
+        """Click the Button control"""
+        self.action.do_action_by_name("click")
+        return self
+
+    # -----------------------------------------------------------
+    def toggle(self):
+        """Method to change toggle button state"""
+        self.click()
```
vasily-v-ryabov: `return self` is also desired here.

pywinauto/controls/atspi_controls.py
```diff
@@ -0,0 +1,345 @@
...
+            return lines[line_index]
+        elif line_index == self.line_count() - 1:
+            return ""
+        else:
+            raise IndexError("There are only {0} lines but given index is {1}".format(self.line_count(), line_index))
+
+    def texts(self):
+        """Get the text of the edit control"""
+        texts = [self.get_line(i) for i in range(self.line_count())]
+        return texts
```
vasily-v-ryabov: Creating a variable is not needed before return statement.

pywinauto/controls/atspi_controls.py
```diff
@@ -0,0 +1,345 @@
...
+        lines = self.window_text().splitlines()
+        if line_index < len(lines):
+            return lines[line_index]
+        elif line_index == self.line_count() - 1:
+            return ""
+        else:
+            raise IndexError("There are only {0} lines but given index is {1}".format(self.line_count(), line_index))
+
+    def texts(self):
+        """Get the text of the edit control"""
```
vasily-v-ryabov: The docstring is the same as for `text_block` method.

pywinauto/controls/atspiwrapper.py
```diff
@@ -0,0 +1,168 @@
...
+          instance or subclass of AtspiWrapper.
+        If the handle is not valid then an InvalidWindowHandle error
+        is raised.
+        """
+        BaseWrapper.__init__(self, element_info, backend.registry.backends['atspi'])
+
+    # ------------------------------------------------------------
+    def set_keyboard_focus(self):
+        """Set the focus to this element"""
+        self.element_info.component.grab_focus("screen")
```
vasily-v-ryabov: `return self` is useful here and below.

pywinauto/keyboard.py
```diff
@@ -1,716 +1,717 @@
...
+            return "KEsc {}".format(self.key)
+
+        def run(self):
+            """Execute the action"""
+            # it works more stable for virtual keys than SendInput
+            for inp in self.GetInput():
+                win32api.keybd_event(inp.ki.wVk, inp.ki.wScan, inp.ki.dwFlags)
+
+
+    class PauseAction():
```
vasily-v-ryabov: Any class shall be inherited from `object` at least.

pywinauto/linux/application.py
```diff
@@ -0,0 +1,171 @@
...
+
+        The action is performed according to only one of parameters
+
+        :param process: a process ID of the target
+        :param path: a path used to launch the target
+
+        .. seealso::
+
+           :func:`pywinauto.findwindows.find_elements` - the keyword arguments that
+           are also can be used instead of **process**, **handle** or **path**
```
vasily-v-ryabov: `handle` is not mentioned above in Linux implementation. It should be removed here as well.

pywinauto/linux/application.py
```diff
@@ -0,0 +1,171 @@
...
+                    continue
+
+                if kwargs['path'] in " ".join(content):
+                    self.process = int(proc_id)
+                    connected = True
+                    break
+
+        if not connected:
+            raise RuntimeError(
+                "You must specify one of process, handle or path")
```
vasily-v-ryabov: `handle` is not used on Linux.

pywinauto/linux/application.py
```diff
@@ -0,0 +1,171 @@
...
+
+    def cpu_usage(self, interval=None):
+        """Return CPU usage percent during specified number of seconds"""
+        if not self.process:
+            raise AppNotConnected("Please use start or connect before trying "
+                                  "anything else")
+        if interval:
+            time.sleep(interval)
+        try:
+            proc_info = subprocess.check_output(["ps", "-p", str(self.process), "-o", "%cpu"], universal_newlines=True)
```
vasily-v-ryabov: I think `%cputime` is more relevant in case of non-None `interval` value. But we have to start `ps` command twice then.

pywinauto/linux/atspi_element_info.py
```diff
@@ -0,0 +1,226 @@
...
+
+    def __hash__(self):
+        """Return a unique hash value based on the element's handle"""
+        return hash(self.handle)
+
+    def __eq__(self, other):
+        """Check if two AtspiElementInfo objects describe the same element"""
+        if self.control_type == "Application":
+            return self.process_id == other.process_id
+        return self.rectangle == other.rectangle
```
vasily-v-ryabov: Maybe add `and self.handle == other.handle` at the end.

pywinauto/linux/atspi_element_info.py
```diff
@@ -0,0 +1,226 @@
...
+            self.__get_elements(el, tree, **kwargs)
+
+    def __hash__(self):
+        """Return a unique hash value based on the element's handle"""
+        return hash(self.handle)
+
+    def __eq__(self, other):
+        """Check if two AtspiElementInfo objects describe the same element"""
+        if self.control_type == "Application":
+            return self.process_id == other.process_id
```
vasily-v-ryabov: This is not correct if `other` is not `Application`.

pywinauto/linux/atspi_element_info.py
airelil: Hmm... It breaks the following test: 
```Python
            desktop = AtspiElementInfo()
            desktop1 = AtspiElementInfo()
            assert desktop = desktop1
```
That's because the constructor creates two different AtspiAccessible objects under the hood.

pywinauto/linux/atspi_element_info.py
```diff
@@ -0,0 +1,232 @@
...
+    @property
+    def rich_text(self):
+        """Return the text of the element"""
+        return self.name
+
+    @property
+    def control_type(self):
+        """Return the class name of the element"""
+        role_id = self.atspi_accessible.get_role(self._handle, None)
+        return IATSPI().known_control_type_ids[role_id]
```
vasily-v-ryabov: It's worth catching `IndexError` and re-raising it as `NotImplementedError('Unknown role ID has been retrieved: ...')`

pywinauto/linux/atspi_objects.py
```diff
@@ -0,0 +1,1225 @@
...
+    ('name', c_char_p),
+    ('description', c_char_p),
+    ('states', POINTER(_AtspiStateSet)),
+    ('attributes', POINTER(_GHashTable)),
+    ('cached_properties', c_uint),
+    ('priv', POINTER(_AtspiAccessiblePrivate)),
+]
+
+_AtspiStateSet._fields_ = [
+    ('fake', c_uint64 * 4),
```
vasily-v-ryabov: These fields can be moved to `_AtspiStateSet` declaration since there is no self-references.

pywinauto/linux/atspi_objects.py
```diff
@@ -0,0 +1,1225 @@
...
+
+class GLIB(object):
+
+    """Python wrapper around C functions from GLib library"""
+
+    LIB12 = "libglib-1.2.so"
+    LIB20 = "libglib-2.0.so"
+    LIB22 = "libglib-2.2.so"
+    DEFAULT_LIB_NAME = "libglib.so"
+    glib = cdll.LoadLibrary(_find_library([LIB12, LIB20, LIB22, DEFAULT_LIB_NAME]))
```
vasily-v-ryabov: I think newer versions should have higher priority: `[LIB22, LIB20, LIB12, DEFAULT_LIB_NAME]`

pywinauto/remote_memory_block.py
```diff
@@ -35,18 +35,18 @@
...
 from ctypes import sizeof
 from ctypes import byref
 from ctypes import c_size_t
 from ctypes import WinError
-from . import win32functions
-from . import win32defines
-from . import win32structures
+import win32api
+
+from .windows import win32functions
```
vasily-v-ryabov: The whole `remote_memory_block.py` should be moved to `windows` folder. It's completely Windows only implementation for "win32" backend only.

pywinauto/controls/atspi_controls.py
```diff
@@ -116,6 +119,7 @@ def is_expanded(self):
         return self.children()[0].is_visible()
 
     def texts(self):
+        """Get texts of all items in in the control as array"""
```
vasily-v-ryabov: Typo: `in in`. I would also prefer saying `as list` which is more correct.

Issue #873, opened 2020-01-01
Organize documentation about Application and WindowSpecification objects better


vasily-v-ryabov:
at the root folder and in the OS specific implementations.

airelil:
As a first step, I can fix the documentation generation in atspi branch

vasily-v-ryabov:
I've merged the first PR. What I could find is that method `start` contains just list of params, but there is no doc string (at the end of page): https://airelil-pywinauto.readthedocs.io/en/atspi/code/pywinauto.application.html

airelil:
Yeah... That's because now we parse the Linux implementation of Application. I'll update the doc strings in `python.linux.applicaiont.Application`, but now wee need to keep it in mind that Linux version of any class will be the main reference in the doc.

PR #881, opened 2020-01-17
issue #873: fix win32 mocks for sphinx docs generation

airelil:
- Update the docs with the new paths of win32-specific modules
- Fix sphinx warnings
- Refactor win32_hooks to allow doc generation on Linux
- Stabilize TaskbarTestCases.testClickVisibleIcon unittest
- atspi backend documentation


pywinauto/windows/win32_hooks.py
```diff
@@ -67,10 +65,12 @@
 
 from .win32defines import VK_PACKET
 from pywinauto.actionlogger import ActionLogger
+from pywinauto.windows import win32functions
```
vasily-v-ryabov: We should use relative imports everywhere. Here: `from . import win32functions` and `from ..actionlogger import ActionLogger`

airelil:
BTW, the preview of the docs for this PR is available here:
https://airelil-pywinauto.readthedocs.io/en/atspi/

Issue #872, opened 2020-01-01
Restore binding WindowSpecification to Application instead of process


vasily-v-ryabov:
It's broken in branch `atspi` (PR #848).

Unit tests must be written to make sure it will not be broken in the future.

PR #841, opened 2019-10-27
WIP: macOS implementation of keyboard, mouse, AxElementInfo and Application

vasily-v-ryabov:
This is initial implementation of the most basic classes for macOS backend "ax".

Any comments and suggestions are highly appreciated.

Big TODOs:
 * Travis CI automation is coming (currently working on macOS 10.11 usage, test apps are built as `.dmg` packages).
 * `AxWrapper` and derived classes (even not started yet).

Thanks:
 * @zemtsov-artem for `mouse` and `keyboard` modules.
 * @briaclg for `AxElementInfo` and `Application` classes.

PR #884, opened 2020-01-24
issue #872: restore binding WindowSpecification to Application

airelil:
Rework linux Application unit tests

### DIFF
--- a/pywinauto/unittests/test_application_linux.py
+++ b/pywinauto/unittests/test_application_linux.py
@@ -8,7 +8,7 @@ import time
 sys.path.append(".")
 from pywinauto.linux.application import Application, AppStartError, AppNotConnected
 from pywinauto.application import WindowSpecification  # noqa: E402
-if sys.platform != 'win32':
+if sys.platform.startswith('linux'):
     from pywinauto.controls import atspiwrapper  # register atspi backend
 
 app_name = r"gtk_example.py"
@@ -26,7 +26,7 @@ def _test_app():
 
 sys.path.append(".")
 
-if sys.platform != 'win32':
+if sys.platform.startswith('linux'):
     class ApplicationTestCases(unittest.TestCase):
 
         """Unit tests for the application.Application class"""
--- a/pywinauto/unittests/test_application_linux.py
+++ b/pywinauto/unittests/test_application_linux.py
@@ -1,14 +1,15 @@
-# TODO crossplatform join this tests with test_application.py
+# TODO crossplatform join these tests with test_application.py
 import sys
 import os
 import unittest
 import subprocess
 import time
-#import pprint
-#import pdb
 
 sys.path.append(".")
 from pywinauto.linux.application import Application, AppStartError, AppNotConnected
+from pywinauto.application import WindowSpecification  # noqa: E402
+if sys.platform != 'win32':
+    from pywinauto.controls import atspiwrapper  # register atspi backend
 
 app_name = r"gtk_example.py"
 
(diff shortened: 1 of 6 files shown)


pywinauto/unittests/test_application_linux.py
```diff
@@ -56,72 +58,75 @@ def test_start_problem(self):
...
+        def setUp(self):
+            """Set some data and ensure the application is in the state we want"""
+            self.app = Application()
+
+        def tearDown(self):
+            """Close the application after tests"""
+            self.app.kill()
+
+        def test_app_binding(self):
+            self.app.start(_test_app())
```
vasily-v-ryabov: Previous `self.app` instance seems to keep running after the test is finished. It's worth preventive killing it.

pywinauto/unittests/test_application_linux.py
```diff
@@ -56,72 +58,75 @@ def test_start_problem(self):
...
+            self.assertEqual(self.app.NonExistingDialog.app, self.app)
+            self.assertEqual(self.app.Application.Panel.exists(), True)
+            self.assertEqual(self.app.Application.Panel.app, self.app)
```
vasily-v-ryabov: A little surprise to me that it works this way. I mean access keywords in the window spec instance by attribute. OK. Just thought to run some real action or check for existence.

Another thought is that WindowSpecification must be created and saved to variable before starting new instance in the same `app` object. Because creating spec after re-start shouldn't catch the issue, should it?

pywinauto/unittests/test_application_linux.py
airelil: Not sure I understand. Do you mean an app instance from the previous test suite: "ApplicationTestCases" ? We have `self.app.kill()` calls in tearDown methods of all test suites in this file.

pywinauto/unittests/test_application_linux.py
airelil: > A little surprise to me that it works this way. I mean access keywords in the window spec instance by attribute.

It was always the case that internal attributes are looked through before going to resolve any external attribute: https://github.com/pywinauto/pywinauto/blob/4005bff2aacb6e5421a43abdbdd2f4c178ca53b4/pywinauto/base_application.py#L365


> Another thought is that WindowSpecification must be created and saved to variable before starting new instance in the same app object.

Apparently, an old instance of WindowSpecification will continue running as far as the app instance is the same. If the app restarts different program the spec will fail to resolve the described control, but if we restart the same program the description is still valid and will be resolved even for a different PID. I think this behavior is just fine. But I'll add another test with Application restart.

pywinauto/unittests/test_application_linux.py
vasily-v-ryabov: You call `self.app.start` twice: in `setUp` method and in the test. But `kill` is called once only. Did you run the tests manually on local machine? Don't you have some non-killed windows?

pywinauto/unittests/test_application_linux.py
```diff
@@ -1,14 +1,15 @@
...
 import unittest
 import subprocess
 import time
-#import pprint
-#import pdb
 
 sys.path.append(".")
 from pywinauto.linux.application import Application, AppStartError, AppNotConnected
+from pywinauto.application import WindowSpecification  # noqa: E402
+if sys.platform != 'win32':
```
vasily-v-ryabov: We shouldn't skip `'darwin'`. Using `.startswith("linux")` is more preferable.

pywinauto/unittests/test_application_linux.py
vasily-v-ryabov: Yes, thanks. But I thought this another test is the core case.

pywinauto/unittests/test_application_linux.py
vasily-v-ryabov: Oops, sorry. `setUp` doesn't call `self.app.start`. My fault.

pywinauto/unittests/test_keyboard.py
```diff
@@ -283,7 +283,7 @@ def testShiftModifier(self):
         received = self.receive_text()
         self.assertEqual("A", received)
 
-    if sys.platform != 'win32':
+    if sys.platform.startswith('linux'):
```
vasily-v-ryabov: Not sure we need all such conditions to change. It was already changed somewhere in macOS branch (see PR #841). So some tests are running on both Linux and macOS with the same code (at least for mouse, keyboard and clipboard). I'd suggest to leave the fix only for the code I've commented on.

vasily-v-ryabov: Maybe change the most recent commit with force push to avoid conflicts in neighbor PR.

PR #885, opened 2020-01-26
Add information about installation with conda to docs and README

### LINKED ISSUE #855: Check "pip install" for conda and maybe prepare conda package using conda package manager
The problem is intially described here:
https://stackoverflow.com/q/58638520/3648361

Also I have old thoughts about checking conda package manager and apt/yum (.deb/.rpm) package for Linux which becomes actual for next major release with "atspi" backend.

m-rossi:
Fix #855.

What‘s not fully clear to me master or atspi branch?


vasily-v-ryabov:
master branch is fine for the actual docs update. `atspi` is for functional changes for next major release. Thanks!

Issue #852, opened 2019-11-21
Revise Application.cpu_usage() for atspi backend


airelil:
The initial implementation of linux.Application.cpu_usage() provides general cpu usage as reported by `ps` linux utility. From `ps` man:
> %cpu - cpu utilization of the process in
> 	 "##.#" format.  Currently, it is
> 	 the CPU time used divided by the
> 	 time the process has been running
> 	 (cputime/realtime ratio), expressed
> 	 as a percentage.  It will not add
> 	 up to 100% unless you are lucky.

We might investigate for a better option to report CPU usage over the specified period.

Issue #547, opened 2018-08-15
uia APP.kill() bug


meshuggahtas:
I'm not sure about this, but my observation is that `APP.kill()` depending on the state of the program throws an error or "halfly terminates" probably because of trying to press ESC on a non existing control/window.

Example 1 : windows magnifier. This only happens, when you click on the dropdown menu AND the window itself was focused.
Example 2 : OBS. This only happens, when the Stats dialog is open AND the OBS window itself was focused.
Example 3 : windows magnifier half termination -> simple example code is `APP = pywinauto.Application(backend="uia").start("magnify.exe")` and `APP.kill()`. If you have docked magnifier, the top of the screen remains "docked". Also I have high confidence in that the application still lingers in the background/memory (huge performance drop while writing this issue), have no idea if this is directly related to pywinatuo `APP.kill()`.

**Example 1 code (Make sure to rename "Nagyito" to Magnifier maybe, name of the window and "Nezetek" to Views maybe, the dropdown with 3 values):**
```
import pywinauto
from pywinauto import actionlogger
actionlogger.enable()

APP = pywinauto.Application(backend="uia").start("magnify.exe")
APP.Nagyito.set_focus() # name of the window maybe Magnifier
APP.Nagyito.Nezetek.click_input() # name of the dropdown maybe Views
APP.kill()
```
**Result of above code:**
```
2018-08-15 11:48:47,341 INFO: Started magnify.exe application.
2018-08-15 11:48:48,710 INFO: Clicked Button "Nézetek" by left button mouse click at (267, 354)
2018-08-15 11:48:49,165 INFO: Typed text to the Menu: {ESC}
Traceback (most recent call last):
  File "C:\Users\oliver.horvath\venv\pywinautogithub\lib\site-packages\pywinauto\uia_defines.py", line 232, in get_elem_interface
    iface = cur_ptrn.QueryInterface(cls_name)
  File "C:\Users\oliver.horvath\venv\pywinautogithub\lib\site-packages\comtypes\__init__.py", line 1158, in QueryInterface
    self.__com_QueryInterface(byref(iid), byref(p))
ValueError: NULL COM pointer access

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "C:\Users\oliver.horvath\venv\pywinautogithub\lib\site-packages\pywinauto\controls\uiawrapper.py", line 434, in close
    iface = self.iface_window
[...]

PR #886, opened 2020-01-31
issue #852, issue #547, typos

airelil:
List of all fixes:
- issue #852: re-work Application.cpu_usage for Linux platform
- typo in ATSPI `ComboBoxWrapper.text` doc string
- move `client_to_screen` implementation into BaseWrapper. This is to have a single method for all backends
- spaces and indentation in linux/application.py
- issue #547: throw same exception from HwndWrapper.close and UIAWrapper.close
    Both backends should raise the same WindowNotFoundError exception if their `close` methods fail to find the window to close

### DIFF
--- a/pywinauto/controls/uiawrapper.py
+++ b/pywinauto/controls/uiawrapper.py
@@ -40,7 +40,7 @@ import warnings
 import comtypes
 
 from .. import backend
-from ..findwindows import WindowNotFoundError
+from .. import WindowNotFoundError  # noqa #E402
 from ..timings import Timings
 from .win_base_wrapper import WinBaseWrapper
 from ..base_wrapper import BaseMeta
@@ -131,7 +131,7 @@ class LazyProperty(object):
         if obj is None:
             return None
         value = self.fget(obj)
-        setattr(obj, self.func_name, value)
+        #setattr(obj, self.func_name, value)
         return value
 lazy_property = LazyProperty
 
--- a/pywinauto/controls/uiawrapper.py
+++ b/pywinauto/controls/uiawrapper.py
@@ -131,7 +131,7 @@ class LazyProperty(object):
         if obj is None:
             return None
         value = self.fget(obj)
-        #setattr(obj, self.func_name, value)
+        setattr(obj, self.func_name, value)
         return value
 lazy_property = LazyProperty
 
--- a/pywinauto/controls/uiawrapper.py
+++ b/pywinauto/controls/uiawrapper.py
@@ -40,6 +40,7 @@ import warnings
 import comtypes
 
 from .. import backend
+from ..findwindows import WindowNotFoundError
 from ..timings import Timings
 from .win_base_wrapper import WinBaseWrapper
 from ..base_wrapper import BaseMeta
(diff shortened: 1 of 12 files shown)


pywinauto/controls/uiawrapper.py
```diff
@@ -40,6 +40,7 @@
 import comtypes
 
 from .. import backend
+from ..findwindows import WindowNotFoundError
```
vasily-v-ryabov: Maybe it's worth using `from .. import WindowNotFoundError` to align with hwndwrapper.py. Not critical. Just to prepare for removing it from `findwindows`.

pywinauto/linux/application.py
```diff
@@ -120,13 +121,33 @@ def cpu_usage(self, interval=None):
...
+                interval = Timings.cpu_usage_interval
+            time.sleep(interval)
+            after = read_cpu_info()
+            pid_time = (after[0] - before[0]) + (after[1] - before[1])
+            sys_time = (after[2] - before[2]) + (after[3] - before[3])
+            if not sys_time:
+                res = 0.0
+            else:
+                res = 100.0 * (float(pid_time) / float(sys_time))
+            #print("linux app cpu usage: ", res)
```
vasily-v-ryabov: Debug print?

pywinauto/linux/application.py
```diff
@@ -120,13 +121,33 @@ def cpu_usage(self, interval=None):
...
+            if not interval:
+                interval = Timings.cpu_usage_interval
+            time.sleep(interval)
+            after = read_cpu_info()
+            pid_time = (after[0] - before[0]) + (after[1] - before[1])
+            sys_time = (after[2] - before[2]) + (after[3] - before[3])
+            if not sys_time:
+                res = 0.0
+            else:
+                res = 100.0 * (float(pid_time) / float(sys_time))
```
vasily-v-ryabov: Maybe I misunderstand something... In my mind `sys_time` is how much time the app spent in the Linux kernel (in dependent kernel modules) and `pid_time` is how much time the app spent in user space modules. So the formula should be a sum of `sys_time` and `pid_time` divided by `interval`. Nope?

airelil: Yes, you got confused with the variable name. Pls, have a look at `read_cpu_info` helper. It returns four values: user time and kernel times for pid and for all system. So the `pid_time` keeps user and kernel time of the process, while `sys_time` stores user time and kernel time of the system. I should rename it to `total_time`

vasily-v-ryabov: OK, now got it.

pywinauto/controls/uiawrapper.py
```diff
@@ -131,7 +131,7 @@ def __get__(self, obj, cls):
         if obj is None:
             return None
         value = self.fget(obj)
-        setattr(obj, self.func_name, value)
+        #setattr(obj, self.func_name, value)
```
vasily-v-ryabov: What is the reason to comment this line?

airelil: It was internal debug accidentally committed.

Issue #911, opened 2020-04-11
w.top_level_parent() raises an exception when w is a "list item" of the Windows start menu


beuaaa:
## Expected Behavior

w.top_level_parent() should not raise a exception when w is a "list item" of the Windows start menu (Like "Google Chrome" or "Notepad++")

Maybe the exception is raised because "semantic zoom" is not a control type known by Pywinauto? 
See the screenshot of inspect.exe in this [link](https://stackoverflow.com/questions/60959635/how-to-get-top-level-parent-of-an-element-in-the-windows-start-menu)

## Actual Behavior

```
Traceback (most recent call last):
  File "C:\Users\d_pra\PycharmProjects\pywinauto_recorder\pywinauto_recorder\recorder.py", line 270, in get_wrapper_path
    wrapper_top_level_parent = wrapper.top_level_parent()
  File "D:\Python\lib\site-packages\pywinauto\base_wrapper.py", line 439, in top_level_parent
    return self.parent().top_level_parent()
  File "D:\Python\lib\site-packages\pywinauto\base_wrapper.py", line 439, in top_level_parent
    return self.parent().top_level_parent()
  File "D:\Python\lib\site-packages\pywinauto\base_wrapper.py", line 433, in top_level_parent
    parent = self.parent()
  File "D:\Python\lib\site-packages\pywinauto\base_wrapper.py", line 409, in parent
    return self.backend.generic_wrapper_class(parent_elem)
  File "D:\Python\lib\site-packages\pywinauto\controls\uiawrapper.py", line 186, in __new__
    return super(UIAWrapper, cls)._create_wrapper(cls, element_info, UIAWrapper)
  File "D:\Python\lib\site-packages\pywinauto\base_wrapper.py", line 126, in _create_wrapper
    new_class = cls_spec.find_wrapper(element_info)
  File "D:\Python\lib\site-packages\pywinauto\controls\uiawrapper.py", line 159, in find_wrapper
    if element.control_type in UiaMeta.control_type_to_cls:
  File "D:\Python\lib\site-packages\pywinauto\uia_element_info.py", line 250, in control_type
    return self._get_control_type()
  File "D:\Python\lib\site-packages\pywinauto\uia_element_info.py", line 112, in _get_current_control_type
    return IUIA().known_control_type_ids[self._element.CurrentControlType]
KeyError: 50039
```

## Steps to Reproduce the Problem

[...]

PR #912, opened 2020-04-11
Update uia_defines.py

beuaaa:
Fixes bug #911 "w.top_level_parent() raises an exception when w is a "list item" of the Windows start menu"


airelil:
Kudos, for making a contribution! Appreciate it. However, as I mentioned in #912 , the control types listshould be populated automatically. The idea was to avoid maintaining a hard-coded list of controls, since it depends on Windows OS and UIAutomation.dll versions installed on a particular machine.

Firstly, I would try to understand why the SemanticZoom control ID is missing in the list.

beuaaa:
Actually, it's better to cancel this pull request.

Issue #668, opened 2019-02-01
Implement WindowWrapper for UIA backend with move_window() method


vasily-v-ryabov:
Found this while reading this StackOverflow question: https://stackoverflow.com/q/54475685/3648361

vasily-v-ryabov:
One more question: https://stackoverflow.com/q/61165628/3648361

PR #922, opened 2020-04-25
Issue #668: add WindowWrapper for uia backend

airelil:
The wrapper contains only `move_window` and `is_dialog` methods.
For others UIA controls the `move_window` throws AttributeError.
Make RECT iterable

### DIFF
--- a/pywinauto/unittests/test_uiawrapper.py
+++ b/pywinauto/unittests/test_uiawrapper.py
@@ -5,6 +5,7 @@ from __future__ import unicode_literals
 import time
 import os
 import sys
+import collections
 import unittest
 import mock
 import six
@@ -18,7 +19,6 @@ from pywinauto.actionlogger import ActionLogger  # noqa: E402
 from pywinauto import Desktop
 from pywinauto import mouse  # noqa: E402
 from pywinauto import WindowNotFoundError  # noqa: E402
-from pywinauto.windows import win32structures  # noqa: E402
 if UIA_support:
     import comtypes
     import pywinauto.windows.uia_defines as uia_defs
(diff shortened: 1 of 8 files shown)


airelil:
Still have Codacy warning as below even though `UIAWrapper.move_window` and `uia_controls.WindowWrapper.move_window` params are identical. The warning could be because of  the additional `repaint` parameter in the original `HwndWrapper.move_window`. I'm thinking about removing the `repaint` from `HwndWrapper.move_window` too. From a quick search through GitHub I didn't see `move_window` calls specifying `repaint` argument explicitly. Seems that everyone relays on the default value.

### [Codacy](https://app.codacy.com/manual/pywinauto/pywinauto/commit?cid=460457689) detected an issue:
#### Message: `Parameters differ from overridden 'move_window' method`

    
#### Currently on:
+ **Commit**:  7d811133d3eba4e21f86f3f82b9ad22e3c82977a
+ **File**:    [pywinauto/controls/uia_controls.py](https://github.com/pywinauto/pywinauto/blob/7d811133d3eba4e21f86f3f82b9ad22e3c82977a/pywinauto/controls/uia_controls.py)
+ **LineNum**: [65](https://github.com/pywinauto/pywinauto/blob/7d811133d3eba4e21f86f3f82b9ad22e3c82977a/pywinauto/controls/uia_controls.py#L65)

pywinauto/unittests/test_uiawrapper.py
```diff
@@ -179,6 +180,45 @@ def test_is_dialog(self):
...
+            self.assertRaises(NotImplementedError, button.move_window)
+
+            dlg_rect = self.dlg.parent().rectangle()  # use the parent as a reference
+            prev_rect = self.dlg.rectangle() - dlg_rect
+
+            new_rect = win32structures.RECT(prev_rect)
+            new_rect.left -= 1
+            new_rect.top -= 1
+            new_rect.right += 2
+            new_rect.bottom += 2
```
vasily-v-ryabov, line 204: Do we really need `win32structures` dependency in UIA tests? I believe it could look like 4-tuple: `(left, top, right, bottom) = prev_rect` but it produces `TypeError: 'RECT' object is not iterable` which should be fixed. `RECT` should be iterable with such a simple conversion to 4-tuple like we have for `POINT`.

Also I'm not sure that so commonly used `class RECT` is in a proper module. Maybe move it out from `win32structures.py`. But I have no idea for a new location yet. Maybe it's not needed right now.

airelil: I added `RECT.__iter__` and opted out to named tuple in the test for now. Also to make linter happy changed the exception for `UIAWrapper.move_window` to AttributeError,

pywinauto/unittests/test_uiawrapper.py
```diff
@@ -8,6 +8,7 @@
 import unittest
 import mock
 import six
+import collections
```
vasily-v-ryabov, line 11: (minor) Built-in module should be imported before 3rd-party one.

dmwyatt:
@vasily-v-ryabov it looks like this isn't in a version released on pypi.  Any chance we'll get a new release anytime soon?

vasily-v-ryabov:
Hi @dmwyatt `atspi` branch has many changes but we plan to make a release in next half a year. No exact deadline yet and guarantee we will fit into half a year. Because it's a hobby project and we missed deadlines many times. Let's how it will go this year.

PR #944, opened 2020-06-10
Fix application.py docstring: "app.child_window" -> "app.window"

stereowind:
Simple fix for a typo in docstring for Application class.  
Example in docstring uses `child_window` method on `app` instance which doesn't exist in `Application` class. Method `window` should be used in this case.  
This got me stumbled a bit while going through documentation and doing hands-on in REPL, hence this PR. I read in another PR that `master` branch should be used for documentation fixes, please correct me if I'm wrong.


airelil:
Can you please re-submit this PR to atspi branch? Notice that application module has neen moved around in atspi branch

PR #987, opened 2020-09-25
More thorough check whether UIA can be used

robertschulze:
In order to prevent failed DLL loads e.g. under WINE


vasily-v-ryabov:
This breaks an idea of loading `UIAutomationCore.dll` in a singleton. Also as I said it is just a workaround for WINE to make "win32" backend working, while "uia" backend is silently skipped. It could lead to side effect on some normal Windows setups.

PR #994, opened 2020-10-18
Add workaround to fix set_focus() failure on UIA (call win32 in this case)

eltimen:
Sometimes set_focus() on UIA backend failed without any exceptions (for example, on Wireshark)

### DIFF
--- a/pywinauto/controls/uiawrapper.py
+++ b/pywinauto/controls/uiawrapper.py
@@ -409,7 +409,10 @@ class UIAWrapper(WinBaseWrapper):
             pass
         try:
             self.element_info.element.SetFocus()
-            if self.element_info != UIAElementInfo.get_active():
+
+            # SetFocus() can return S_OK even if the element isn't focused actually
+            active_element = UIAElementInfo.get_active()
+            if self.element_info != active_element and self.element_info != active_element.top_level_parent:
                 if self.handle:
                     warnings.warn("Failed to set focus on element, trying win32 backend", RuntimeWarning)
                     HwndWrapper(self.element_info).set_focus()
--- a/pywinauto/controls/uiawrapper.py
+++ b/pywinauto/controls/uiawrapper.py
@@ -43,6 +43,7 @@ from .. import backend
 from .. import WindowNotFoundError  # noqa #E402
 from ..timings import Timings
 from .win_base_wrapper import WinBaseWrapper
+from .hwndwrapper import HwndWrapper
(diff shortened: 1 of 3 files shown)


eltimen:
@vasily-v-ryabov, I just discovered that condition https://github.com/pywinauto/pywinauto/blob/1aecf7347603af74a4fb4d07a01d7ab7debb9091/pywinauto/controls/uiawrapper.py#L412
can be true if the `set_focus` worked fine. For example, in script
```
app = Application(backend="uia").connect(path=r"notepad.exe")
win = app.window(best_match='UntitledNotepad')
win.set_focus()
```
`SetFocus()` from UIA works (the window becomes foreground), but it still switches to win32 backend.
The reason for this is that children widget grabs focus automatically:
`self.element_info` returns `uia_element_info.UIAElementInfo - 'Untitled - Notepad', Notepad` and
`UIAElementInfo.get_active()` returns `uia_element_info.UIAElementInfo - '', Edit`.

I tried to fix it with the following code, but it doesn't work for Wireshark:
```
if self.element_info.element.SetFocus() != S_OK:
```

vasily-v-ryabov:
There is method `top_level_parent()` in `BaseWrapper`. I think we can move it to base class `ElementInfo` (to avoid duplication in all backends) so that `.top_level_parent` property uses backend-specific `.parent` property. The mentioned condition would look like `if self.element_info != UIAElementInfo.get_active() and self.element_info != UIAElementInfo.get_active().top_level_parent:` Maybe you need to save `UIAElementInfo.get_active()` result to a variable so it's not called twice.

pywinauto/controls/uiawrapper.py
```diff
@@ -408,9 +409,23 @@ def set_focus(self):
...
+                                  "the window doesn't have native handle", RuntimeWarning)
         except comtypes.COMError as exc:
-            warnings.warn('The window has not been focused due to ' \
-                'COMError: {}'.format(exc), RuntimeWarning)
+            if self.handle:
+                warnings.warn("Failed to set focus on element due to COMError: {}, "
+                              "trying win32 backend".format(exc), RuntimeWarning)
+                HwndWrapper(self.element_info).set_focus()
+            else:
+                warnings.warn("The window has not been focused due to COMError: {} "
```
vasily-v-ryabov, line 426: I'd suggest to put comma after `{}` for all such warnings to separate COMError message from the rest of warning text.

vasily-v-ryabov (approved):

Issue #696, opened 2019-03-27
Can i select toolbar option using pywinauto


Nagesh1504:
Hi..
i have question how to select tool bar option using pywinauto. please find below snap shot of the application which has tool option,

![image](https://user-images.githubusercontent.com/48903263/55067157-36c1b080-50a5-11e9-9a4a-760cda02f9dc.png)

i am trying select 'Settings' option but i could not able to select. i have tried below code.
app.WaveMaster.Settings.click()

but i get error like,
Traceback (most recent call last):
  File "C:\Python27\Test_1_1.py", line 61, in <module>
    print app.Settings.Button.click()
  File "C:\Python27\lib\site-packages\pywinauto\application.py", line 352, in __getattribute__
    ctrls = self.__resolve_control(self.criteria)
  File "C:\Python27\lib\site-packages\pywinauto\application.py", line 249, in __resolve_control
    raise e.original_exception
MatchError: Could not find 'Settings' in '[u'WaveMasterDialog', u'Dialog', u'WaveMaster']'

Please help me on this,

Nagesh1504:
Hi all,

i also tried with below code while selecting 'Settings' option from above snapshot.

1. app.WaveMaster.Settings.SETTINGSButton.click_input()

got plenty of error, 
Traceback (most recent call last):
  File "C:\Python27\Test_1_1.py", line 61, in <module>
    app.WaveMaster.Settings.SETTINGSButton.click_input()
  File "C:\Python27\lib\site-packages\pywinauto\application.py", line 352, in __getattribute__
    ctrls = self.__resolve_control(self.criteria)
  File "C:\Python27\lib\site-packages\pywinauto\application.py", line 249, in __resolve_control
    raise e.original_exception
[...]

Issue #644, opened 2019-01-02
Qt5: dump_tree() cannot list "detouched" children of main window for Dolphin Emulator, but Inspect.exe can


vasily-v-ryabov:
Found on StackOverflow: https://stackoverflow.com/q/54003752/3648361

There is Dolphin Emulator downloads: https://dolphin-emu.org/download/

Latest version 5.0-9281 has interesting app side issue that Inspect.exe can workaround. Main window property "Children" getter returns this error (listed in Inspect.exe):
```
[Error: calling getter for this property: hr=0xFFFFFFFF80004005 - Unspecified error]
```
But Inspect.exe shows the hierarchy with "[+]" button anyway. This would be interesting to investigate and apply the same workaround in pywinauto.

vasily-v-ryabov:
I think we can close this issue after enabling optional way to use `FindAll()`.

eltimen:
> ```
> [Error: calling getter for this property: hr=0xFFFFFFFF80004005 - Unspecified error]
> ```
I tried to reproduce this, and the above error message doesn't show in my environment (Windows 7 x64, UIAutomationCore.dll version 7.0.0.0).

During debugging the `print_control_identifiers()` problem, I found out that COM method `IUIAutomationElement::FindAll` raises `COMError (-2147024882, 'Not enough memory resources are available to complete this operation.', (None, None, None, 0, None))`. Details known to me:
1. This problem only occurs when FindAll called with `TreeScope_Descendants` (inside `print_control_identifiers()` and `descendants()`), but `children()` method works fine.
2. The problem is probably not related to the width or depth of the widget tree (because it works fine with Qt-based benchmark with 62 nested widgets).
3. The thread `UIAutomationCore.dll!UiaHTextRangeFromVariant` is created inside `Dolphin.exe` process after FindAll call and causes high CPU usage. Probably, the root of the problem can be found by examining stack of this thread: <details> <summary>Screenshot</summary>
![screen](https://user-images.githubusercontent.com/53598853/98091986-36031a00-1e97-11eb-9f1b-e15c2aee2839.jpg)
</details>

I have not found other cases where this bug appears, besides Dolphin. But the problem is solved after replacing `FindAll` with `RawViewWalker`.

eltimen:
The possible reason of all this oddities is recursion inside elements tree:

![изображение](https://user-images.githubusercontent.com/53598853/101280799-2e0f0080-37dc-11eb-94a6-950bf2c49ae4.png)

PR #1001, opened 2020-10-29
Change UIA backend to use RawViewWalker instead of FindAll

eltimen:
Methods `children()` and `descendants()` of the UIAElementInfo changed to use `IUIAutomation::RawViewWalker` instead of `IUIAutomationElement::FindAll`.

This allows to find more elements than FindAll returns (for example, solves the problem mentioned in #644)

Related issues #644 #696

### DIFF
--- a/pywinauto/unittests/test_uiawrapper.py
+++ b/pywinauto/unittests/test_uiawrapper.py
@@ -1613,7 +1613,7 @@ if UIA_support:
             """Test selecting a WPF menu item by index"""
             path = "#0->#1->#1"  # "File->Close->Later"
             self.dlg.menu_select(path)
-            label = self.dlg.MenuLaterClickLabel.wrapper_object()
+            label = self.dlg.MenuLaterClickStatic.wrapper_object()
             self.assertEqual(label.window_text(), u"MenuLaterClick")
 
             # Non-existing paths
@@ -1626,7 +1626,7 @@ if UIA_support:
             """Test selecting a WPF menu item by exact text match"""
             path = "File->Close->Later"
             self.dlg.menu_select(path, True)
-            label = self.dlg.MenuLaterClickLabel.wrapper_object()
+            label = self.dlg.MenuLaterClickStatic.wrapper_object()
             self.assertEqual(label.window_text(), u"MenuLaterClick")
 
             # A non-exact menu name
@@ -1637,14 +1637,14 @@ if UIA_support:
             """Test selecting a WPF menu item by best match text"""
             path = "file-> close -> later"
             self.dlg.menu_select(path, False)
-            label = self.dlg.MenuLaterClickLabel.wrapper_object()
+            label = self.dlg.MenuLaterClickStatic.wrapper_object()
             self.assertEqual(label.window_text(), u"MenuLaterClick")
 
         def test_menu_by_mixed_match(self):
             """Test selecting a WPF menu item by a path with mixed specifiers"""
             path = "file-> #1 -> later"
             self.dlg.menu_select(path, False)
-            label = self.dlg.MenuLaterClickLabel.wrapper_object()
+            label = self.dlg.MenuLaterClickStatic.wrapper_object()
             self.assertEqual(label.window_text(), u"MenuLaterClick")
 
             # Bad specifiers
--- a/pywinauto/unittests/test_uiawrapper.py
+++ b/pywinauto/unittests/test_uiawrapper.py
@@ -73,17 +73,11 @@ if UIA_support:
         def test_issue_296(self):
             """Test handling of disappered descendants"""
             wrp = self.dlg.wrapper_object()
-            orig = uia_defs.IUIA().raw_tree_walker.GetFirstChildElement
-            uia_defs.IUIA().raw_tree_walker.GetFirstChildElement = mock.Mock(
-                side_effect=ValueError("Mocked value error"),
-                return_value=[])  # empty list
-            self.assertEqual([], wrp.descendants())
-            exception_err = comtypes.COMError(-2147220991, "Mocked COM error", ())
-            uia_defs.IUIA().raw_tree_walker.GetFirstChildElement = mock.Mock(
-                side_effect=exception_err,
-                return_value=[])  # empty list
-            self.assertEqual([], wrp.descendants())
-            uia_defs.IUIA().raw_tree_walker.GetFirstChildElement = orig  # restore the original method
+            with mock.patch.object(uia_defs.IUIA().raw_tree_walker, 'GetFirstChildElement') as mock_get_first_child:
+                mock_get_first_child.side_effect = ValueError("Mocked value error")
+                self.assertEqual([], wrp.descendants())
+                mock_get_first_child.side_effect = comtypes.COMError(-2147220991, "Mocked COM error", ())
+                self.assertEqual([], wrp.descendants())
 
         def test_issue_278(self):
             """Test that statement menu = app.MainWindow.Menu works for 'uia' backend"""
--- a/pywinauto/unittests/test_uiawrapper.py
+++ b/pywinauto/unittests/test_uiawrapper.py
@@ -73,15 +73,17 @@ if UIA_support:
         def test_issue_296(self):
             """Test handling of disappered descendants"""
             wrp = self.dlg.wrapper_object()
-            orig = wrp.element_info._element.FindAll
-            wrp.element_info._element.FindAll = mock.Mock(side_effect=ValueError("Mocked value error"),
-                                                          return_value=[])  # empty list
+            orig = uia_defs.IUIA().raw_tree_walker.GetFirstChildElement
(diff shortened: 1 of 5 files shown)


eltimen:
I tried to compare the performance of both methods and got the following results:

Test application: modified *widgets\mainwindows\mainwindow* example from Qt5 with number of dock windows and nested groupboxes specified via command line arguments

**Window with 7 children elements, average tree depth 2**

|Function|FindAll median time (seconds)|UIA median time (seconds)|
|---------|---------|------------|
|children()|0.115|0.191|
|descendants(depth=3)|0.699|0.297|
|descendants()|0.211|0.390|

**Window with 12 children elements, average tree depth 7**

|Function|FindAll median time (seconds)|UIA median time (seconds)|
|---------|---------|------------|
|children()|0.113|0.220|
|descendants(depth=3)|0.979|0.329|
|descendants()|0.217|0.774|

**Window with 67 children elements, average tree depth 62**

|Function|FindAll median time (seconds)|UIA median time (seconds)|
|---------|---------|------------|
|children()|0.127|0.367|
|descendants(depth=3)|22.766|1.285|
|descendants()|3.384|74.173|

Source of used benchmarks located [here](https://github.com/eltimen/pywinauto/blob/perf/perftests/benchmark.py)

pywinauto/windows/uia_element_info.py
```diff
@@ -55,6 +56,28 @@ def elements_from_uia_array(ptrs, cache_enable=False):
     return elements
 
 
+def is_element_satisfies_criteria(element, process=None, class_name=None, name=None, control_type=None,
```
airelil, line 59: Grammar nitpicking, the name should be "is_element_satisfying_criteria" or "element_satisfies_criteria"

pywinauto/windows/uia_element_info.py
```diff
@@ -55,6 +56,28 @@ def elements_from_uia_array(ptrs, cache_enable=False):
...
+                                  content_only=None, **kwargs):
+    """Check if element satisfy filter criteria"""
```
airelil, line 61: Also should be:
"""Check if element satisfies filter criteria"""

pywinauto/windows/uia_element_info.py
```diff
@@ -55,6 +56,28 @@ def elements_from_uia_array(ptrs, cache_enable=False):
...
+    is_appropriate_control_type = True
+    if control_type:
+        if isinstance(control_type, string_types):
+            is_appropriate_control_type = element.CurrentControlType == IUIA().known_control_types[control_type]
+        elif not isinstance(control_type, int):
```
airelil, line 66: If in element_info.py above already changed to ` if isinstance(depth, integer_types) `, shouldn't use the same check against `six.integer_types`?

pywinauto/windows/uia_element_info.py
```diff
@@ -280,52 +303,70 @@ def parent(self):
...
-        """Find all elements according to the given tree scope and conditions"""
+    # def _get_elements(self, tree_scope, cond=IUIA().true_condition, cache_enable=False):
+    #     """Find all elements according to the given tree scope and conditions"""
+    #     try:
+    #         ptrs_array = self._element.FindAll(tree_scope, cond)
+    #         return elements_from_uia_array(ptrs_array, cache_enable)
+    #     except(COMError, ValueError) as e:
+    #         print(e)
+    #         ActionLogger().log("COM error: can't get elements")
+    #         return []
```
airelil, line 314: I think, the unused code should be removed, instead of just commenting it of.

pywinauto/windows/uia_element_info.py
```diff
@@ -280,52 +303,70 @@ def parent(self):
...
+            if is_element_satisfies_criteria(element, **kwargs):
+                yield UIAElementInfo(element, cache_enable)
 
-    def iter_children(self, **kwargs):
-        """Return a generator of only immediate children of the element
+    def iter_descendants(self, **kwargs):
+        """Iterate over descendants of the element"""
+        cache_enable = kwargs.pop('cache_enable', False)
+        depth = kwargs.pop("depth", None)
+        if not isinstance(depth, (int, type(None))) or isinstance(depth, int) and depth < 0:
```
airelil, line 341: Same here, should stick to `six.integer_types` or I'm missing something?

pywinauto/windows/uia_element_info.py
```diff
@@ -280,52 +303,70 @@ def parent(self):
...
+            raise Exception("Depth must be natural number")
```
airelil, line 342: Probably: "Depth must be an integer"

pywinauto/unittests/test_uiawrapper.py
```diff
@@ -73,15 +73,17 @@ def tearDown(self):
...
             self.assertEqual([], wrp.descendants())
             exception_err = comtypes.COMError(-2147220991, "Mocked COM error", ())
-            wrp.element_info._element.FindAll = mock.Mock(side_effect=exception_err,
-                                                          return_value=[])  # empty list
+            uia_defs.IUIA().raw_tree_walker.GetFirstChildElement = mock.Mock(
+                side_effect=exception_err,
+                return_value=[])  # empty list
             self.assertEqual([], wrp.descendants())
-            wrp.element_info._element.FindAll = orig  # restore the original method
+            uia_defs.IUIA().raw_tree_walker.GetFirstChildElement = orig  # restore the original method
```
airelil, line 86: Honestly, we should switch to a mock context manager, because the original method won't be restored if test asserts and it may affect outcome of other tests.

airelil (changes requested):
Good fix. I just had a quick go through.

pywinauto/windows/uia_element_info.py
eltimen: done

pywinauto/windows/uia_element_info.py
eltimen: done

pywinauto/windows/uia_element_info.py
eltimen: Change in element_info.py is caused by a failed test on python 2.7. I agree, it's useful to make same changes here just in case.

pywinauto/windows/uia_element_info.py
eltimen: done

pywinauto/windows/uia_element_info.py
eltimen: Fixed, it will be better

pywinauto/unittests/test_uiawrapper.py
eltimen: Hmm, I tried to write this:
```
wrp = self.dlg.wrapper_object()
with mock.patch.object(uia_defs.IUIA().raw_tree_walker, 'GetFirstChildElement') as mock_get_first_child:
    mock_get_first_child = mock.Mock(
        side_effect=ValueError("Mocked value error"),
        return_value=[])  # empty list
    self.assertEqual([], wrp.descendants())
```
and got exception on `isinstance` check, caused by replacement to `MagicMock` object:
```
File "C:\repo\pywinauto\github\pywinauto\pywinauto\windows\uia_element_info.py", line 337, in iter_descendants
    yield UIAElementInfo(child, cache_enable)
File "C:\repo\pywinauto\github\pywinauto\pywinauto\windows\uia_element_info.py", line 113, in __init__
    "with integer or IUIAutomationElement instance only!")
TypeError: UIAElementInfo object can be initialized with integer or IUIAutomationElement instance only!
```

I will try to find a solution to this problem.

eltimen: done at https://github.com/pywinauto/pywinauto/pull/1001/commits/99a2b300ac5675d305222b9aa09b7c709db735dc

pywinauto/windows/uia_element_info.py
airelil: I think so, if we stop using it now.

pywinauto/windows/uia_element_info.py
vasily-v-ryabov: `integer` -> `an integer`

pywinauto/windows/uia_element_info.py
vasily-v-ryabov: We haven't decided it finally. Maybe it's worth to keep this code to allow enabling it as an option. Also it's useful for performance comparisons. Maybe it will lead to some trade-offs between `FindAll` and `RawViewWalker` approaches.

vasily-v-ryabov (approved):

vasily-v-ryabov:
@eltimen thanks for measurements! Please create a Wiki page with these numbers and put a link on it here.

eltimen:
> 
> 
> @eltimen thanks for measurements! Please create a Wiki page with these numbers and put a link on it here.

https://github.com/pywinauto/pywinauto/wiki/Performance-measurements-(draft)

airelil (approved):

PR #1009, opened 2020-11-22
Update hook_and_listen.py

daaawx:
I've fixed the imports to make the code run without errors.

### DIFF
--- a/examples/hook_and_listen.py
+++ b/examples/hook_and_listen.py
@@ -1,9 +1,7 @@
 import win32api
 import win32con
 from threading import Timer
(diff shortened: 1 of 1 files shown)


vasily-v-ryabov:
Hi @daaawx please note that current active branch is `atspi` while current release branch (0.6.8) is `master`. So you've made a "fix" for master branch, but created a PR to `atspi` branch. This is wrong because `master` contains correct example `hook_and_listen.py` which works with release version of pywinauto. Sorry for confusion. We accept PRs to `atspi` branch only since it is being prepared for next major release 0.7.0. We have no plans for 0.6.x updates any more.

Issue #807, opened 2019-09-05
Normalize terminology?


pakal:
As a newcomer to GUI automation, I was quite confused by the terminology used in methods : for me "window" always meant top-level window/popup (that with the minimize/close buttons), and sub-entities were called "controls" or "widgets".

In the current API, window/child_window are used to designate all kinds of frames/widgets/controls, and some similarly-named functions (eg. several def windows()) tweak top_level_only/visible_only/enabled_only parameters in a way which is not obvious to API users, or mask some possibilities (eg. Application.windows() might actually be used to fetch any widgets, by adding top_level_only=False).

I read there was a redesign in progress, is their any plan to rename some methods like this (while still leaving compatibility aliases of course) ?

vasily-v-ryabov:
Leaving compatibility aliases is not a must since we're working on new major release with all these re-designs. We want to break backward compatibility once and reduce the most of confusion.

Candidates to remove or redesign:
 * `.top_window()` is a leader of confusion since it returns window specification by handle at the moment of calling. So when new window becomes top, the returned value is not changed automatically so it refers to previous old window that may even not exist any more.
 * `.window()/.child_window()` should be renamed to `.by()`. WindowSpecification is similar to Locator concept in Selenium, but I have no idea how to name it without confusion and without copying name "locator".

All boolean properties are already cut and designed for 3 possible values (on a branch with 2 non-fixed tests left):
 * `visible=True` means "visible only"
 * `visible=False` means "invisible only"
 * `visible=None` means "all" (no filtering by this property, default value for the vast majority of cases).

Do you have any more ideas?

vasily-v-ryabov:
`top_level_only` will be replaced by `depth` keyword (integer number of levels to dive into).

pakal:
Sounds good, WindowSpecification could become WidgetSpecification or WidgetLocator I guess ?

I'm worried about backwards compatibility breakages however; from what I've seen with python2to3 and other major "cleanups", the effects can be dreadful on the library's ecosystem. If, unlike my own case, enterprises and free projects have big codebases of pywinauto tests, they will almost never allocate the numerous days necessary to rewritting them to conform to the new API (it's already rare to obtain worlkoad to migrate production code...); and it will create a schism between new and legacy users, the latter either staying on an obsolete version of pywinauto, and/or developing new test suites on a whole new basis (appium, katalon, ranorex...).

[...]

Issue #710, opened 2019-04-12
Implement WindowSpecification.find_all() and .find() methods instead of .wrapper_object()


bernd-wechner:
I was using WindowSpecification.child_window (singular) to find windows with great success. I cmae across a messy problem where it bombed because of an ambiguity (more than one window matching the criteria). Try as I might with print_control_identifiers() I could not find them, though in the end did. Cost me some time and energy mind you. And much of that would have been saved if rather than this bombing:

`open_dialog = my_window.window(title_re=name, control_type="Pane")`

with an Ambiguity, I could quickly and simply change it to:

`open_dialog = my_window.windows(title_re=name, control_type="Pane")`

And get the list to work with!

Which I have done the hard way, but racing down and replicating the call to find_elements. The thing that was bothering me and tool me some while to grasp was that parent= takes not a wrapped element but its element_info ... and that find_element needs a parent and backend specified, the defaults being applied by things that call it. So as per its documentation, it's low level and down and dirty ;-). 

Now it turns out this software has some interest control structure, in that it has two identically named Panes one the parent of the other and the only differentiator I can find is the class_name of the element. Everything else appears identical bar the parent and children() properties. One it seems is a docking control, and the other the actual Pane (it being wrapped in a docking control that allows you to move it around and dock it in the interface).

Still, finding this duplicity would be MUCH easier if the above recommendation was in place, a simple way to move from a Ambiguity Exception to a returned list to examine. I need to do this now anyhow with find_elements to choose one of such a pair based on the relationship between them).

airelil:
Cheers on finding your way with ```find_elements```! As a quick note. For top level windows we already have ```Application.windows()``` and ``` Desktop.windows()``` methods. And if you already have a wrapper object,  you can use ```descendants``` or the recent generator-based method ```iter_descendants```. For "uia" backend, these methods accept several criterias, such as process, class_name, control_type, title and content_only. Would it suit your needs?

vasily-v-ryabov:
This is also part of nearest re-design. It would look so: `.by(name=name, control_type="Pane").find_all()` or `.find()` at your choice.

Currently it's possible to try several indices: `.child_window(title=name, control_type="Pane", found_index=0)`, then `found_index=1` etc. This is kinda workaround so far. This param won't be necessary after code re-factoring.

pakal:
This feature would be marvelous  :)

[...]

PR #1016, opened 2020-12-06
Code refactoring in WindowsSpecification

EOKruglov:
Linked issues: 
* https://github.com/pywinauto/pywinauto/issues/710
* https://github.com/pywinauto/pywinauto/issues/807

Changes:
* rename .wrapper_object() to .find(); wrapper_object marked as deprecated; WrapperObject was removed.
* rename .child_window() to .by(); child_window, window marked as deprecated ChildWindow, Window, Window_, window_ were removed
* Exists, Wait, WaitNot, PrintControlIdentifiers were removed

### DIFF
--- a/apps/Gtk_samples/gtk_example.py
+++ b/apps/Gtk_samples/gtk_example.py
@@ -37,7 +37,7 @@ def _add_image_widget():
     img.set_tooltip_text("orange image")
     return img
 
-class TestApplicationMainWindow(Gtk.by):
+class TestApplicationMainWindow(Gtk.Window):
 
     def _add_combobox(self):
         country_store = Gtk.ListStore(str)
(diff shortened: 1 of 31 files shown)


apps/Gtk_samples/gtk_example.py
```diff
@@ -100,7 +100,7 @@ def _create_textview(self):
         return scrolledwindow
 
     def __init__(self):
-        Gtk.Window.__init__(self, title="MainWindow")
+        Gtk.by.__init__(self, title="MainWindow")
```
vasily-v-ryabov, line 103: Looks like wrong replace. This is not connected with the core library code.

sandbox/recorder.py
```diff
@@ -84,8 +84,8 @@ def OnKeyboardEvent(event):
 
     app = Application()
 
-    if event.Window:
-        wrapped = app.window(handle = event.Window)
+    if event.by:
+        wrapped = app.window(handle = event.by)
```
vasily-v-ryabov, line 88: This file shouldn't be touched as well.

vasily-v-ryabov (changes requested):

vasily-v-ryabov (approved):

vasily-v-ryabov:
`import warnings` is not used any more in `base_application.py`. So it can be removed.

EOKruglov:
> `import warnings` is not used any more in `base_application.py`. So it can be removed.

removed

PR #1047, opened 2021-03-01
Implement method get_active() (fix #760), add Qt5 combo box workaround (fix #876)

### LINKED ISSUE #760: Implement HwndWrapper get_child_focus() method
This request has been reported on StackOverflow:
https://stackoverflow.com/q/56774363/3648361

Now method `.has_keyboard_focus()` uses `win32gui.GetFocus()` to get keyboard focused element for check purpose. But there is no method that returns such element.

### LINKED ISSUE #876: Unrelated combobox is operated when select a item of combobox
## Expected Behavior
An item in the combobox is selected.

## Actual Behavior
After correct behavior, an unrelated combobox is changed and expanded.

## Short Example of Code to Demonstrate the Problem
I tried the following source by the NewControls in [VCSamples](https://github.com/microsoft/VCSamples).
`VCSamples/VC2010Samples/MFC/Visual C++ 2008 Feature Pack/NewControls/`

newcontrol

### source
~~~py
# launch the app
app = Application("uia").start(r"..\newcontrols.exe")
dlg = app.window(title_re = "NewControls")
# get combobox with resource ID
combobox = dlg.child_window(auto_id = '1002')
# select a combobox item
combobox.select('Image on right')
~~~

In this example, I try to change only the 'Image Location" combobox, but the "Appearance" combobox is operated and expanded without collapsing..

### note
* I used `backend='uia'`, because I want to use resource ID of MFC.
* When changing the source to select with number (ex.`combobox.select(2)`), the source cannot work.
* After removing source under `# do health check and apply workaround for Qt5 combo box if necessary` in `select()` method of `ComboBoxWrapper` class, the program worked fine.

## Specifications
  - Pywinauto version: 0.6.8
  - Python version and bitness: Python3.7(64bit)
  - Platform and OS: Windows 10

This is the first time I have written to the github.
So, if there is a problem in my writing, I would be glad if you could point out.

RadaevValentin:


pywinauto/unittests/test_application.py
```diff
@@ -1393,6 +1401,11 @@ def test_get_list_of_windows_win32(self):
         window_titles = [win_obj.window_text() for win_obj in dlgs]
         self.assertTrue(self.window_title in window_titles)
 
+    def test_issue_760_desktop_win32(self):
+        focused_element = self.desktop.get_active()
+        if not isinstance(focused_element, HwndWrapper):
+            raise Exception("Received object is not an instance of HwndWrapper")
```
vasily-v-ryabov, line 1407: Better way is using `self.assertEqual(type(focused_element), HwndWrapper)`. It is one line, and it will print which exact type in case of failure. Please do it for all such tests.

apps/QT5_Combobox/QT5_combobox.py
```diff
@@ -0,0 +1,37 @@
...
+def window():
+    app = QApplication(sys.argv)
+    win = MyWindow()
+    win.show()
+    win.setAccessibleName("QTRV")
+    win.show()
+
+    sys.exit(app.exec_())
+
+window()
```
vasily-v-ryabov, line 37: Please add empty line at end of file.

apps/QT5_Combobox/QT5_combobox.py
```diff
@@ -0,0 +1,37 @@
+from PyQt5 import QtWidgets
+from PyQt5.QtWidgets import QApplication, QMainWindow
+
+import sys
```
vasily-v-ryabov, line 4: Move `import sys` before 3rd-party imports.

pywinauto/unittests/test_application.py
```diff
@@ -47,6 +47,9 @@
 import mock
 import six
 
+from pywinauto.controls.hwndwrapper import HwndWrapper
+from pywinauto.controls.uiawrapper import UIAWrapper
```
vasily-v-ryabov, line 51: If you need these wrappers, maybe move these tests to `test_hwndwrapper.py` and `test_uiawrapper.py`? It shouldn't add new dependent modules.

apps/QT5_Combobox/QT5_combobox.py
```diff
@@ -0,0 +1,37 @@
+from PyQt5 import QtWidgets
```
vasily-v-ryabov, line 1: All Python files should be named according to PEP8: i.e. `qt5_combobox.py`.

vasily-v-ryabov (changes requested):
In general it looks better.

 * But please make normal commit messages, don't be too short. If you want to refer an issue in commit message, use `#` sign. For example, `#876` will look like #876. Also make commit message as a normal sentence: `Fix combobox issue #<issue_number>` or something like that.
 * Also please use correct commit email that matches to your profile or bind  both emails to your GitHub profile.
 * And please take a look at test failures on AppVeyor (see "Checks" expandable section at the end of pull request).

pywinauto/unittests/test_application.py
vasily-v-ryabov: It also breaks some tests because backend specific tests have check if comtypes is installed. Some runs have UIA_SUPPORT=NO (comtypes not installed). It is designed specifically to detect such architecture issues.

dev-requirements.txt
```diff
@@ -7,4 +7,5 @@ codecov
 rst2pdf
 Sphinx
 mock==2.0.0
-codacy-coverage
\ No newline at end of file
+codacy-coverage
+PyQt5
```
vasily-v-ryabov, line 11: Looks like PyQt5 package doesn't exist (not maintained?) for Python2.7. It's possible to add conditional record to requirements file. It's easy to google. Also if you don't use this app yet, you may not add PyQt5 to requirements right now.

vasily-v-ryabov: https://stackoverflow.com/q/19559247/3648361

pywinauto/unittests/test_application.py
```diff
@@ -1302,7 +1302,6 @@ def test_folder_list(self):
                              [u'x64', u'BCDialogMenu.exe', u'CmnCtrl1.exe', u'CmnCtrl2.exe', u'CmnCtrl3.exe',
                               u'CtrlTest.exe', u'mfc100u.dll', u'RebarTest.exe', u'RowList.exe', u'TrayMenu.exe'])
             self.assertEqual(files_list.item('RebarTest.exe').window_text(), 'RebarTest.exe')
-
```
vasily-v-ryabov, line 1305: Empty line between methods shouldn't be removed.

vasily-v-ryabov (commented):
Now looks good. Just need to amend commit history.

pywinauto/unittests/test_hwndwrapper.py
```diff
@@ -246,6 +251,10 @@ def testTopLevelParent(self):
         self.assertEqual(self.ctrl.top_level_parent(), self.dlg.handle)
         self.assertEqual(self.dlg.top_level_parent(), self.dlg.handle)
 
+    def test_issue_760_desktop_win32(self):
```
vasily-v-ryabov, line 254: Issue number should be in test name only if it is a bug. Name it `test_get_active_desktop` by feature name. The same for base wrapper.

apps/QT5_Combobox/qt5_combobox.py
```diff
@@ -0,0 +1,40 @@
...
+from PyQt5 import QtWidgets
+from PyQt5.QtWidgets import QApplication, QMainWindow
+
+
+class MyWindow(QMainWindow):
+    def __init__(self):
+        super(MyWindow, self).__init__()
+        self.initUI()
+
+    def init_ui(self):
```
vasily-v-ryabov, line 12: PyQt5 expects method `initUI`.

pywinauto/unittests/test_application.py
```diff
@@ -1326,7 +1326,7 @@ def test_folder_list(self):
                              [u'x64', u'BCDialogMenu.exe', u'CmnCtrl1.exe', u'CmnCtrl2.exe', u'CmnCtrl3.exe',
                               u'CtrlTest.exe', u'mfc100u.dll', u'NewControls.exe', u'RebarTest.exe', u'RowList.exe', u'TrayMenu.exe'])
             self.assertEqual(files_list.item('RebarTest.exe').window_text(), 'RebarTest.exe')
-
+
```
vasily-v-ryabov, line 1329: Please remove trailing spaces

pywinauto/controls/uia_controls.py
```diff
@@ -303,25 +303,37 @@ def select(self, item):
...
                 # do health check and apply workaround for Qt5 combo box if necessary
                 if isinstance(item, six.string_types):
-                    item = children_lst[0].children(name=item)[0]
-                    if self.selected_text() != item:
-                        # workaround for WinForms combo box
-                        item.invoke()
-                        if self.selected_text() != item:
-                            # workaround for Qt5 combo box
+                    item = list_view.children(name=item)[0]
+                    item_value = item.texts()[0]
```
vasily-v-ryabov, line 311: Why not getting `item.window_text()`? It gets only its own text instead of all children.

pywinauto/controls/uia_controls.py
```diff
@@ -356,7 +368,11 @@ def selected_index(self):
             return self.selected_item_index()
         except NoPatternInterfaceError:
             # workaround for Qt5 and WinForms
-            return self.texts().index(self.selected_text())
+            try:
+                return self.children(control_type='List')[0].texts().index(self.selected_text().splitlines())
+            except IndexError:
```
vasily-v-ryabov, line 373: The above line contains 2 places which can raise `IndexError`. It looks a little bit dangerous to catch both. Is it what you want?

PR #1080, opened 2021-05-26
[Howto] A few hints on how to integrate pywinauto and Jenkins

cowo78:
Added a "Integration with Jenkins" chapter to quickly sum up the steps required to get pywinauto integrated into a Jenkins pipeline.

PR #1081, opened 2021-05-28
Added references to PAExec as alternative to PsExec

cowo78:
See comments in #1080


vasily-v-ryabov:
Thanks a lot Giuseppe!

PR #1088, opened 2021-06-20
typo

smalinux:


vasily-v-ryabov:
Please re-direct this PR to active branch `atspi` instead of master.

Issue #555, opened 2018-08-27
dict_keys() is empty when trying to select a ListBox item


vnabokit:
I try to select an item in a ListBox (pywinauto 0.6.5).

The initial script just finds the certain drop down menu inside a dialog.
```
from pywinauto.application import Application
app = Application(backend='uia').start(r"app.exe")
dlg = app.window(best_match='Select Rename Device')
dlg.window(best_match='DownComboBox3').print_control_identifiers()
```
The output seems to be good:
```
ComboBox - ' Down'    (L795, T522, R1157, B542)
['ComboBox', ' DownComboBox', ' Down']
child_window(title=" Down", control_type="ComboBox")
   | 
   | ListBox - ''    (L0, T0, R0, B0)
   | ['', 'ListBox']
   |    | 
   |    | ListItem - '00/4i Simulation mode'    (L1, T1, R752, B24)
   |    | ['00/4i Simulation mode', '00/4i Simulation modeListItem', 'ListItem', 'ListItem0', 'ListItem1']
   |    | child_window(title="00/4i Simulation mode", control_type="ListItem")
   |    | 
   |    | ListItem - '00/2i Simulation mode'    (L1, T24, R752, B47)
   |    | ['00/2i Simulation mode', '00/2i Simulation modeListItem', 'ListItem2']
   |    | child_window(title="00/2i Simulation mode", control_type="ListItem")
   |    | 
   |    | ListItem - '00/6i Simulation mode'    (L1, T47, R752, B70)
   |    | ['00/6i Simulation mode', '00/6i Simulation modeListItem', 'ListItem3']
   |    | child_window(title="00/6i Simulation mode", control_type="ListItem")

Process finished with exit code 0
```

Now I try to select '`00/2i Simulation mode`' ListItem:
```
from pywinauto.application import Application
app = Application(backend='uia').start(r"app.exe")
dlg = app.window(best_match='Select Rename Device')
dlg.window(best_match='DownComboBox3').child_window(best_match='00/2i Simulation mode').select()
```
but got an error:
`pywinauto.findbestmatch.MatchError: Could not find '00/2i Simulation mode' in 'dict_keys([])'`

I tried to apply workaround suggested in the issue #365 but the error was the same:
```
from pywinauto.application import Application
app = Application(backend='uia').start(r"app.exe")
dlg = app.window(best_match='Select Rename Device')
[...]

PR #1128, opened 2021-10-08
Fix ComboBoxWrapper.select() for Qt5 widgets

eltimen:
Also add PyQt5 sample to unit tests.
related to #555 

Note: 
Attempt to start PyQt5 app by this code 
```python3
qt_python_folder = os.path.join(os.path.dirname(__file__), r"..\..\apps\Qt_Python_samples")
qt_py_combobox_app = sys.executable + ' ' + os.path.join(qt_python_folder, u"qt5_combobox.py")

app = Application(backend='uia').start(qt_py_combobox_app)
```
causes exception:
```
  File "pywinauto\pywinauto\unittests\test_uiawrapper.py", line 1579, in setUp
    app = Application(backend='uia').start(qt_py_combobox_app)
  File "pywinauto\pywinauto\windows\application.py", line 446, in start
    if wait_for_idle and not app_idle():
  File "pywinauto\pywinauto\windows\application.py", line 432, in app_idle
    result = win32event.WaitForInputIdle(
pywintypes.error: (2, 'WaitForInputIdle', 'The system cannot find the file specified.')
```
Workarounded with
```python3
app = Application(backend='uia').start(qt_py_combobox_app, wait_for_idle=False)
self.app = Application(backend='uia').connect(pid=app.process)
```

### DIFF
--- a/pywinauto/controls/uia_controls.py
+++ b/pywinauto/controls/uia_controls.py
@@ -330,7 +330,12 @@ class ComboBoxWrapper(uiawrapper.UIAWrapper):
                 elif self.selected_index() != item:
                     items = children_lst[0].children(control_type='ListItem')
                     if item < len(items):
-                        items[item].invoke()
+                        if self.element_info.framework_id == 'Qt':
+                            list_view._select(item)
+                            if list_view.is_active():
+                                items[item].click_input()
+                        else:
+                            items[item].invoke()
                     else:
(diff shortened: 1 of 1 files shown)


vasily-v-ryabov:
`wait_for_idle=False` is OK because `python.exe` starts console first which has no GUI event loop. So this param is made for such cases exactly.

PR #1135, opened 2021-10-18
Handle 87 error from OpenProcess and CI changes

### LINKED ISSUE #1127: Handle error 87 (The parameter is incorrect) in all methods containing call to OpenProcess()
This is reported on StackOverflow: https://stackoverflow.com/q/69342624/3648361

Error 87 should be handled by error code, because error message can be translated to another language. It means process doesn't exist any more. Maybe such error can be caught in other methods as well. So we can raise more clear `ProcessNotFoundError` in this case.

eltimen:
resolves #1127
I think another `OpenProcess` calls without winapi errors handling (inside `WaitGuiThreadIdle` and `RemoteMemoryBlock` used by `HwndElementInfo` and various control wrappers) do not require handling of this error.

CI changes:
* Change python version matrix (remove some jobs, add Python 3.8)
* Fix `pywin32 is in an unsupported or invalid wheel` on Python 3.6

### DIFF
--- a/ci/runTestsuite.ps1
+++ b/ci/runTestsuite.ps1
@@ -42,7 +42,7 @@ function run {
     $output = "transformed.xml"
     
     #nosetests  --all-modules --with-xunit pywinauto/unittests
-    nosetests --nologcapture --exclude=testall --with-xunit --with-coverage --cover-html --cover-html-dir=Coverage_report --cover-package=pywinauto --verbosity=3 pywinauto\unittests
+    nosetests --nologcapture --traverse-namespace --exclude=testall --with-xunit --with-coverage --cover-html --cover-html-dir=Coverage_report --cover-package=pywinauto --verbosity=3 pywinauto\unittests
     $success = $?
     Write-Host "result code of nosetests:" $success
 
--- a/ci/runTestsuite.ps1
+++ b/ci/runTestsuite.ps1
@@ -42,6 +42,7 @@ function run {
     $output = "transformed.xml"
     
     #nosetests  --all-modules --with-xunit pywinauto/unittests
+    # --traverse-namespace is required for python 3.8 https://stackoverflow.com/q/58556183
     nosetests --nologcapture --traverse-namespace --exclude=testall --with-xunit --with-coverage --cover-html --cover-html-dir=Coverage_report --cover-package=pywinauto --verbosity=3 pywinauto\unittests
     $success = $?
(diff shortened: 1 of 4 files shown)


pywinauto/windows/application.py
```diff
@@ -478,7 +478,14 @@ def cpu_usage(self, interval=None):
         if not self.process:
             raise AppNotConnected("Please use start or connect before trying "
                                   "anything else")
-        h_process = win32api.OpenProcess(win32con.MAXIMUM_ALLOWED, 0, self.process)
+
+        try:
+            h_process = win32api.OpenProcess(win32con.MAXIMUM_ALLOWED, 0, self.process)
+        except win32api.error as e:
+            if e.winerror == win32defines.ERROR_INVALID_PARAMETER:
+                raise ProcessNotFoundError('Process does not exist')
```
vasily-v-ryabov, line 486: It's better to add PID to error message.

pywinauto/windows/application.py
```diff
@@ -478,7 +478,14 @@ def cpu_usage(self, interval=None):
...
+            else:
```
vasily-v-ryabov, line 487: `else:` is not necessary.

ci/runTestsuite.ps1
```diff
@@ -42,7 +42,7 @@ function run {
     $output = "transformed.xml"
     
     #nosetests  --all-modules --with-xunit pywinauto/unittests
-    nosetests --nologcapture --exclude=testall --with-xunit --with-coverage --cover-html --cover-html-dir=Coverage_report --cover-package=pywinauto --verbosity=3 pywinauto\unittests
+    nosetests --nologcapture --traverse-namespace --exclude=testall --with-xunit --with-coverage --cover-html --cover-html-dir=Coverage_report --cover-package=pywinauto --verbosity=3 pywinauto\unittests
```
vasily-v-ryabov, line 45: It's better to add comment with reference to https://stackoverflow.com/q/58556183/3648361

pywinauto/windows/application.py
eltimen: done

pywinauto/windows/application.py
eltimen: done

ci/runTestsuite.ps1
eltimen: done

vasily-v-ryabov (approved):

Issue #1126, opened 2021-10-08
Implement method item_by_path(path) and maybe others for uia_controls.ToolbarWrapper


vasily-v-ryabov:
This is suggested in Gitter chat (Russian room). We have such method for `uia_controls.MenuWrapper`. But some menus are implemented as toolbar with drop-down submenus. It makes sense to add similar methods to choose such submenu items.

PR #1203, opened 2022-04-13
item_by_path(path) for uia_controls.ToolbarWrapper

EOKruglov:
Issue: [1126](https://github.com/pywinauto/pywinauto/issues/1126)

Changes:
* Added item_by_path(path) method for ToolbarWrapper. One of the buttons on the toolbar is called `DropDownButton`, and can contain `MenuItem` objects, so the method works the same as `MenuWrapper.item_by_path(path)`.
* Fixed a bug, that caused `ToolbarWrapper.buttons()` to return a list with duplicate buttons.


pywinauto/controls/uia_controls.py
```diff
@@ -1212,10 +1216,7 @@ def texts(self):
     def button_count(self):
         """Return a number of buttons on the ToolBar"""
         if self.win32_wrapper is not None:
-            btn_count = self.win32_wrapper.button_count()
-            if btn_count:
-                return btn_count
-            return len(self.win32_wrapper.children())
+            return len(self.buttons())
```
vasily-v-ryabov, line 1219: This is less optimal, because `self.buttons()` walks all buttons and creates a list, but we need only count!

pywinauto/controls/uia_controls.py
```diff
@@ -1302,6 +1307,84 @@ def check_button(self, button_identifier, make_checked, exact=True):
         self.actions.logSectionEnd()
         return button
 
+    # -----------------------------------------------------------
+    def items(self):
+        return self.buttons()
```
vasily-v-ryabov, line 1312: It's better to do `items = buttons` to avoid duplicating docstring etc.

pywinauto/controls/uia_controls.py
```diff
@@ -1302,6 +1307,84 @@ def check_button(self, button_identifier, make_checked, exact=True):
...
+
+    # -----------------------------------------------------------
+    def _activate(self, item, is_last):
+        """Activate the specified item"""
+        if not item.is_active():
+            item.set_focus()
+        try:
+            item.expand()
+        except(NoPatternInterfaceError):
```
vasily-v-ryabov, line 1321: This is different catching style than everywhere in the project. Use space and no rounds brackets.

vasily-v-ryabov (changes requested):

pywinauto/controls/uia_controls.py
EOKruglov: This is kind of a workaround because `win32_wrapper.button_count()` takes into account duplicate buttons

vasily-v-ryabov: Ah, OK. Then please add a comment about it.

pywinauto/controls/uia_controls.py
```diff
@@ -1302,6 +1308,111 @@ def check_button(self, button_identifier, make_checked, exact=True):
...
+
+        def next_level_menu(parent_menu, item_name, is_last):
+            if item_name.startswith("#"):
+                return self._sub_item_by_idx(parent_menu, int(item_name[1:]), is_last)
+            else:
+                return self._sub_item_by_text(parent_menu, item_name, exact, is_last)
+
+        try:
+            if items_cnt == 1:
+                menu = next_level_menu(self, toolbar_items[0], items_cnt == 1)
```
vasily-v-ryabov, line 1388: `items_cnt == 1` is always `True` at third argument (Codacy warning). Maybe use `True` explicitly.

pywinauto/controls/uia_controls.py
```diff
@@ -1302,6 +1308,111 @@ def check_button(self, button_identifier, make_checked, exact=True):
...
+                return menu
+            else:
```
vasily-v-ryabov, line 1390: `else:` is not necessay after `return` statement (Codacy warning). Remove else and de-indent the below code block.

pywinauto/controls/uia_controls.py
```diff
@@ -1302,6 +1308,111 @@ def check_button(self, button_identifier, make_checked, exact=True):
...
+                        new_descendants.append(menu)
+                        try:
+                            for ctrl in new_descendants[::-1]:
+                                try:
+                                    menu = next_level_menu(ctrl, toolbar_items[i], items_cnt == i + 1)
+                                except AttributeError:
+                                    pass
+                        except (findwindows.ElementNotFoundError,
+                                findbestmatch.MatchError):
+                            raise IndexError()
```
vasily-v-ryabov, line 1407: It would be nice to provide more details: which text or index is not found in which list of texts/children?

pywinauto/controls/uia_controls.py
```diff
@@ -1302,6 +1308,111 @@ def check_button(self, button_identifier, make_checked, exact=True):
...
+                    descendants_after = self.top_level_parent().descendants()
+                    new_descendants = list(set(descendants_after) - set(descendants_before))
+
+        except AttributeError:
+            raise IndexError()
```
vasily-v-ryabov, line 1412: Which lines are supposed to fail with such `AttributeError`? Can we make more detailed error message based on this info?

pywinauto/controls/uia_controls.py
```diff
@@ -38,7 +38,7 @@
 from . import uiawrapper
 from . import win32_controls
 from . import common_controls
-from .. import findbestmatch
+from .. import findbestmatch, findwindows
```
vasily-v-ryabov, line 41: This import fails on Py2.7 on AppVeyor (restart shows it's reproducible). Please debug and fix.

vasily-v-ryabov (changes requested):

pywinauto/controls/uia_controls.py
EOKruglov: I suggest to remove this exception. If `items_count==1`, we will get a detailed error message from `findbestmatch.find_best_match`. We can leave one exception in case when `items_count > 1`. What do you think?

pywinauto/controls/uiawrapper.py
```diff
@@ -775,6 +775,24 @@ def menu_select(self, path, exact=False, ):
...
+        The full path syntax is specified in:
+        :py:meth:`pywinauto.controls.uia_controls.ToolbarWrapper.item_by_path`
+        """
+        self.verify_actionable()
+
+        cc = self.children(control_type="ToolBar")
+        if not cc:
+            cc = self.descendants(control_type="ToolBar")
+            if not cc:
+                raise AttributeError
```
vasily-v-ryabov, line 792: Add some message about empty list of buttons/items/children.

pywinauto/controls/uia_controls.py
```diff
@@ -188,6 +188,10 @@ def click(self):
         # Return itself so that action can be chained
         return self
 
+    # -----------------------------------------------------------
+    def items(self):
```
vasily-v-ryabov, line 192: Please add a docstring.

pywinauto/controls/uiawrapper.py
```diff
@@ -775,6 +775,24 @@ def menu_select(self, path, exact=False, ):
...
+        The full path syntax is specified in:
+        :py:meth:`pywinauto.controls.uia_controls.ToolbarWrapper.item_by_path`
+        """
+        self.verify_actionable()
+
+        cc = self.children(control_type="ToolBar")
+        if not cc:
+            cc = self.descendants(control_type="ToolBar")
+            if not cc:
+                raise AttributeError("Can not find ToolBar item in {}".format(self.descendants()))
```
vasily-v-ryabov, line 792: Calling any wrapper method inside `raise SomeException` statement is risky! Because it may raise another exception and add more confusion. Please use general phrase like `"Cannot find any item with control_type=\"ToolBar\" in children and descendants"`.

pywinauto/controls/uia_controls.py
```diff
@@ -1302,6 +1309,108 @@ def check_button(self, button_identifier, make_checked, exact=True):
...
+            - "Tools->#0->Configure"
+
+        * **path** - Path to the specified item. **Required**.
+        * **exact** - If false, text matching will use a 'best match' fuzzy algorithm. If true, will try to find the
+                      item with the given name. (Default False). **Optional**
+        """
+        toolbar_items = [p.strip() for p in path.split("->")]
+        items_cnt = len(toolbar_items)
+        if items_cnt == 0:
+            raise IndexError()
```
vasily-v-ryabov, line 1376: Add message like `"Empty path is not accepted by the method"`

vasily-v-ryabov (changes requested):
Windows tests are still failing with Py2.7 64-bit with UIA. Please fix it!

vasily-v-ryabov (approved):

PR #1218, opened 2022-06-04
atspi test scenario update

RadaevValentin:
methods name aligning with Windows analogues

### DIFF
--- a/pywinauto/controls/atspi_controls.py
+++ b/pywinauto/controls/atspi_controls.py
@@ -496,27 +496,27 @@ class ScrollBarWrapper(atspiwrapper.AtspiWrapper):
     def __init__(self, elem):
         """Initialize the control"""
         super(ScrollBarWrapper, self).__init__(elem)
-        self.value = self.element_info.get_value()
+        self.atspi_value_obj = self.element_info.get_atspi_value_obj()
 
     # -----------------------------------------------------------
-    def get_min_value(self):
+    def min_value(self):
         """Get the minimum value of the ScrollBar"""
-        return self.value.get_minimum_value()
+        return self.atspi_value_obj.get_minimum_value()
 
     # -----------------------------------------------------------
-    def get_max_value(self):
+    def max_value(self):
         """Get the maximum value of the ScrollBar"""
-        return self.value.get_maximum_value()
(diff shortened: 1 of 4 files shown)


pywinauto/controls/atspi_controls.py
```diff
@@ -499,22 +499,22 @@ def __init__(self, elem):
...
 
     # -----------------------------------------------------------
-    def get_min_step(self):
+    def min_step(self):
         """Get the minimum step of the ScrollBar"""
         return self.value.get_minimum_increment()
 
     # -----------------------------------------------------------
-    def get_current_value(self):
+    def value(self):
```
vasily-v-ryabov, line 517: It conflicts with line 499! Attribute is overridden.

vasily-v-ryabov: Also `get_something` is a typical style in other backends. I don't see any reason to remove only some `get_` prefixes. Please keep it as `get_value` etc.

vasily-v-ryabov (changes requested):
This is from Codacy warnings.

pywinauto/controls/atspi_controls.py
vasily-v-ryabov: `def value(self)` is present only in `uia_controls.SliderWrapper`. Let's keep this one. And use `get_value` in other wrappers.

pywinauto/controls/atspi_controls.py
```diff
@@ -496,27 +496,27 @@ class ScrollBarWrapper(atspiwrapper.AtspiWrapper):
     def __init__(self, elem):
         """Initialize the control"""
         super(ScrollBarWrapper, self).__init__(elem)
-        self.value = self.element_info.get_value()
+        self.control_value = self.element_info.get_value()
```
vasily-v-ryabov, line 499: First I would name it `self.atspi_value_obj` to avoid confusion with text value as a search property.

pywinauto/controls/atspi_controls.py
vasily-v-ryabov: Still not renamed back to `get_value`. Why are you not reading the comments?

vasily-v-ryabov (changes requested):

PR #1322, opened 2023-08-14
small fix in `uia_defines`

junkmd:
## delint and rename variables within `get_elem_interface`
In reality, what is passed to `QueryInterface` is not the "class name" but the "pattern class", so I changed the variable names to be appropriate.

## fix base name typo
Previously, it was 'MulipleView' instead of 'MultipleView', which is why `IUIAutomationMultipleViewPattern` or `UIA_MultipleViewPatternId` could not be found in `comtypes.gen.UIAutomationClient`.


vasily-v-ryabov (approved):

junkmd:
If this PR gets merged, I'm planning to work for refactoring `uia_defines`.

junkmd:
@vasily-v-ryabov 
I have rebased this branch onto the current master.
Is this ready to be merged?

PR #1323, opened 2023-08-14
fix `is_checked` in `uia_controls`

junkmd:
I noticed that the `IUIAutomationTogglePattern` assigned to `iface_toggle` does not have an attribute named `ToggleState_On`.

https://learn.microsoft.com/en-us/windows/win32/api/uiautomationclient/nn-uiautomationclient-iuiautomationtogglepattern#methods

`uia_controls.FooWrapper.is_checked` have only been tested for cases where `NoPatternInterfaceError` raises, and there is no test that actually verifies if it is "toggled".

However, I believe it's not ideal to leave a situation where `AttributeError` is consistently occurring unaddressed, so I made the necessary corrections.

### DIFF
--- a/pywinauto/controls/uia_controls.py
+++ b/pywinauto/controls/uia_controls.py
@@ -758,7 +758,7 @@ class ListItemWrapper(uiawrapper.UIAWrapper):
         Only items supporting Toggle pattern should answer.
         Raise NoPatternInterfaceError if the pattern is not supported
         """
-        return self.iface_toggle.ToggleState_On == toggle_state_on
+        return self.iface_toggle.CurrentToggleState == toggle_state_on
 
     def texts(self):
         """Return a list of item texts"""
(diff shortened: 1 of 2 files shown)


vasily-v-ryabov (commented):
Thanks! May I ask for adding such positive test case?

junkmd:
On the other hand, in `ButtonWrapper`, there is a `get_toggle_state` that returns `iface_toggle.CurrentToggleState`, and since the tests are passing, I am confident that `CurrentToggleState` is definitely a functioning property present in `IUIAutomationTogglePattern`.

https://github.com/pywinauto/pywinauto/blob/bf7f789d01b7c66ccd0c213db0a029da7e588c9e/pywinauto/controls/uia_controls.py#L159-L173

https://github.com/pywinauto/pywinauto/blob/bf7f789d01b7c66ccd0c213db0a029da7e588c9e/pywinauto/unittests/test_uiawrapper.py#L740-L755

junkmd:
I thought preparing an actual GUI would incur significant testing costs.
Therefore, I added tests for `is_checked` using the `patch` approach, similar to what's done in other test cases in `test_uiawrapper.py`.

PR #1325, opened 2023-08-19
a much more concise check for the necessary preconditions for UIA

junkmd:
Please see https://github.com/pywinauto/pywinauto/issues/957#issuecomment-1646614400 and https://github.com/pywinauto/pywinauto/issues/957#issuecomment-1646680752


vasily-v-ryabov (approved):
👍

PR #1398, opened 2024-08-16
Update CI pipeline to revive this project.

junkmd:
I want to revive this project, which has been stagnant in maintenance, so I am proposing the following changes:
- Drop support for Python 3.7 and earlier versions
- Add support for the latest versions of Python
- Remove `nosetests`, which no longer works with newer versions of Python
- Introduce a new library for test reporting

The artifacts generated by `unittest` + `coverage` differ from those generated by `nose`, so even if they are uploaded to AppVeyor, the number of tests executed will not be displayed as before.
I thought that if the artifacts were in x-unit format, they might be more suitable for AppVeyor, so I tried using [`unittest-xml-reporting`](https://pypi.org/project/unittest-xml-reporting/), but it didn’t work. 

For now, I have removed the artifact upload from the CI pipeline.
But even if it doesn’t work well with AppVeyor, there may still be value in uploading the report files, so I have no objections to reinstating the artifact upload.

The warnings from Codacy are caused by `Write-Host`.
Since they are already being used elsewhere, I believe there are not problems.


junkmd:
@vasily-v-ryabov 
@airelil 

Could you review this PR?
I would like to help revive this project.
If you lack the resources to dedicate to this project, I am happy to allocate my resources and become one of the maintainers.

vasily-v-ryabov:
`master` branch is incorrect. I planned next major release from branch `atspi`, but this is a little bit complicated. What I plan soon is releasing 0.6.9 as last version supporting Python 2.7 with pinned dependencies like `comtypes<=1.2.1` and `pywin32<=228`. It should look like 1 commit and release. Then we can fully drop Python 3.7- for `atspi` branch and refresh almost everything.

junkmd:
@vasily-v-ryabov 

Thank you for your reply.

> What I plan soon is releasing 0.6.9 as last version supporting Python 2.7 with pinned dependencies like `comtypes<=1.2.1` and `pywin32<=228`. It should look like 1 commit and release.

No objection.

I understand that the 0.6.9 minor version will consist of the current commits on the `master` branch plus "1 commit".

I propose releasing the 0.6.10 minor version by merging this PR after the 0.6.9 release.
As committed [here](https://github.com/pywinauto/pywinauto/pull/1398/commits/72ef03bcf4c37620ef442b421715daa57c594a46), there are parts of the current `master` branch that still use features removed in newer versions of Python, so I believe this step is necessary.
After that, I suggest merging `master` into the `atspi` branch.

junkmd:
The CI pipeline was restored with the merging of #1423.

PR #1345, opened 2023-09-29
fix `ci/install.ps1` and other ci configs for Windows

junkmd:
From the AppVeyor CI results in a previous PR, I noticed that in environments with Python 3.9 and later, `Test-Path $env:PYTHON` in `ci/install.ps1` returns False, and it falls back to Python 3.4.

As a result, even in cases where `UIA_SUPPORT=YES`, `comtypes` is not actually installed, and the functionality of `backend='uia'` is not being tested.

I'm experimenting with various solutions to fix this.

<details>
<summary>trial and error commits</summary>

### make tests runnable for each Python versions

> I tried [removing the fallback conditional](https://github.com/pywinauto/pywinauto/pull/1345/commits/97bd990c1186e72c10d36fd6e0cc27a34fc6fdbb), and while [it worked on Python 3.8](https://ci.appveyor.com/project/pywinauto/pywinauto/builds/48154955/job/lvsv6vnq1uybxjdk?fullLog=true#L7) and earlier, but [it failed on Python 3.9](https://ci.appveyor.com/project/pywinauto/pywinauto/builds/48154955/job/e5e7v0owexy3yv9y?fullLog=true#L7) and later.
> - (Is `C:\\Python3x\Scripts\pip.exe` truly missing from the env?)
> 

68204d2ba5dcf63f0739955cb95bebba4ae92c53
77e9f3d4a20dac98f880b16b2c4f92b151475ba2
6ec3cc90d393d02b16f568d8c8c8be08804cd392
a493bec8edb78ba4235d0676c9db2fb81a659488
c99a2771069409740d9bf714d1f1172cd81716dc
c1dfaa5c3b016dc875f57ccc28b66be4c710a12a
0a1c669fd187dbc963dcb56ab11add817c755a55
1077bd92ed8fe6d63460acfc55cf46f93579fa54
2b2b7a875e1be51ab3347cea564bcdaed0c4704d
83a69a61692b9f8a7e1ab49446adc406342de55d
5f9e12394396c5d4ab02f3bbec5c5e1821ef1eb0
1af2a72aee0d5ac2b2f4e07bf967509305ff8547
663b23b4577202e6278211da985eb62a24a141fa
d529000ad40b117bfca49f6d7f4885a4f7f0092e
f37d82eef85f360b537f0c79b2a522ed8e119551
47d4f4babd3f44a9f5e66e06abf10ca3472d8fbd
115b38d90919445d9d7165300f9066af34d69d93
[...]

Issue #858, opened 2019-11-28
[pytest integration] (Windows fatal exception: code 0x80010108) @ comtypes


ykangpark:
## Description
Running pytest with pywinauto causes an exception in shutdown of comtypes.
Tests are collected / run correctly, but at the end there's always an exception output.

This breaks test discovery in VSCode, though I can work around it by moving import statements to the inside of fixtures / functions.

I am not sure if this is an issue with how I am using pytest and pywinauto, or its a bug.

I tried python 3.6 (Visual Studio 2017), 3.7 (Anaconda3), and 3.8 (python.org).

## Example code
File test_case.py
```
import pytest
from pywinauto.application import Application

def test_case():
    assert True
```

## Test discovery
In the directory containing test_case.py
`python -m pytest --collect-only`

Output
```
========================================================================================= test session starts ========================================================================================== 
platform win32 -- Python 3.8.0, pytest-5.3.1, py-1.8.0, pluggy-0.13.1
rootdir: C:\Users\Yujin\Desktop\pytest_pywinauto
collected 1 item
<Module test_case.py>
  <Function test_case>

=========================================================================================== warnings summary =========================================================================================== 
C:\Users\Yujin\AppData\Roaming\Python\Python38\site-packages\win32\lib\pywintypes.py:2
  C:\Users\Yujin\AppData\Roaming\Python\Python38\site-packages\win32\lib\pywintypes.py:2: DeprecationWarning: the imp module is deprecated in favour of importlib; see the module's documentation for alternative uses
    import imp, sys, os

-- Docs: https://docs.pytest.org/en/latest/warnings.html
========================================================================================== 1 warning in 0.17s ========================================================================================== 
Windows fatal exception: code 0x80010108

Thread 0x0000625c (most recent call first):
  File "C:\Users\Yujin\AppData\Roaming\Python\Python38\site-packages\comtypes\__init__.py", line 185 in shutdown
Windows fatal exception: code 0x80010108

Thread 0x0000625c (most recent call first):
  File "C:\Users\Yujin\AppData\Roaming\Python\Python38\site-packages\comtypes\__init__.py", line 185 in shutdown
```

## Running the test
`python -m pytest`
```
C:\Users\Yujin\Desktop\pytest_pywinauto>"C:/Program Files/Python38/python.exe" -m pytest
========================================================================================= test session starts ========================================================================================== 
[...]

Issue #790, opened 2019-08-12
UIAWrapperTests.test_scroll failure


airelil:
## Expected Behavior
UIAWrapperTests.test_scroll should pass for all configurations

## Actual Behavior
Recently the unittest UIAWrapperTests.test_scroll started failing on AppVeyor
```
Traceback (most recent call last):
  File "C:\projects\pywinauto\pywinauto\controls\uia_controls.py", line 812, in cell
    cell_elem = uiawrapper.UIAWrapper(elem_info)
  File "C:\projects\pywinauto\pywinauto\controls\uiawrapper.py", line 186, in __new__
    return super(UIAWrapper, cls)._create_wrapper(cls, element_info, UIAWrapper)
  File "C:\projects\pywinauto\pywinauto\base_wrapper.py", line 126, in _create_wrapper
    new_class = cls_spec.find_wrapper(element_info)
  File "C:\projects\pywinauto\pywinauto\controls\uiawrapper.py", line 159, in find_wrapper
    if element.control_type in UiaMeta.control_type_to_cls:
  File "C:\projects\pywinauto\pywinauto\uia_element_info.py", line 250, in control_type
    return self._get_control_type()
  File "C:\projects\pywinauto\pywinauto\uia_element_info.py", line 112, in _get_current_control_type
    return IUIA().known_control_type_ids[self._element.CurrentControlType]
ValueError: NULL COM pointer access

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "c:\python35\lib\unittest\case.py", line 59, in testPartExecutor
    yield
  File "c:\python35\lib\unittest\case.py", line 605, in run
    testMethod()
  File "C:\projects\pywinauto\pywinauto\unittests\test_uiawrapper.py", line 250, in test_scroll
    cell = listview.cell(3, 0)
  File "C:\projects\pywinauto\pywinauto\controls\uia_controls.py", line 814, in cell
    raise IndexError
IndexError
```
There is also a failure of TreeViewAdditionalTestCases.testCheckBoxes, but it seems to be fixed with:
https://github.com/airelil/pywinauto/commit/cd9967a885219e67e9c65d0071f3fc934f9333e3

## Specifications

  - Pywinauto version: master branch (and same on my atspi branch)
  - Python version and bitness: 2.7, 3.5, 3.6, 3.7 (x86 and x64)

airelil:
I can't reproduce it locally on my win8.1. Might worth trying on win10...

[...]

PR #1423, opened 2025-01-02
Prepare release 0.6.9 (last Py2.7 compatible one)

vasily-v-ryabov:
This is the last release for Python 2.7 and <=3.6. Next major release `0.7.0` will support 3.7+ only.

The dependencies are also pinned to old versions available for Python 2.7.


junkmd:
Happy new year and thanks to your great works!

I think `test_uiawrapper.UIAWrapperTests.test_scroll` should be marked with `@unittest.skip` (temporarily), as it fails as reported in #790.

The failure of `test_common_controls.TreeViewAdditionalTestCases.testCheckBoxes` is likely due to a timing issue.
In #1398, I addressed this problem by making changes such as switching `Timings.fast` to `Timings.slow`.
Feel free to cherry-pick the commits from my #1398 and add them to your PR.

vasily-v-ryabov:
@junkmd thanks for references and Happy New Year!
It seems I made the very minimal fixes for both CI issues. I could recall that selecting a tab by `.set_focus()` is less reliable (I need a separate fix for `.set_focus()` in 0.7.0), so for this minor release I just used method `.select()` instead.

junkmd:
@vasily-v-ryabov 

The fix for `test_scroll` and the migration to `pytest` are excellent.

Regarding concerns with `pytest`, such as those mentioned in https://github.com/pywinauto/pywinauto/issues/858 and https://github.com/enthought/comtypes/issues/204, it seems there are no problems in this project’s test suite with `faulthandler` or `Windows fatal exception`s.
(Perhaps the need for `-p no:faulthandler` arises in cases like my closed-source project, where multiple plugins are installed for pytest or the production codebase has more complex dependencies.)

BTW, for version 0.7.0, will it only include changes to the CI pipeline and packaging configuration?
If so, will the `atspi` branch be merged into version 0.8.0 or a later major release?

ci/runTestsuite.ps1
```diff
@@ -1,54 +1,37 @@
...
-    $output = "transformed.xml"
-    
-    #nosetests  --all-modules --with-xunit pywinauto/unittests
-    nosetests --nologcapture --exclude=testall --with-xunit --with-coverage --cover-html --cover-html-dir=Coverage_report --cover-package=pywinauto --verbosity=3 pywinauto\unittests
-    $success = $?
-    Write-Host "result code of nosetests:" $success
+    $results = "results.xml"
 
-    xslt_transform $input $stylesheet $output
+    pytest --junit-xml=$results --tb=native --capture=sys --show-capture=no -v --verbosity=3 --cache-clear --durations=15 --ignore=testall.py --log-level=DEBUG --cov-report html:Coverage_report --cov=pywinauto pywinauto\unittests
```
junkmd, line 30: I have confirmed that the `Windows fatal exception` [did not occur in Python 3.6](https://ci.appveyor.com/project/pywinauto/pywinauto/builds/51265079/job/4pilwp2n88u7bc8v) or earlier but [does occur in Python 3.7](https://ci.appveyor.com/project/pywinauto/pywinauto/builds/51265079/job/6qmmf9k4a786rixj) and later when using the combination of `pytest` and `pywinauto`.

Given that this issue does not arise with the standard library's `unittest` and considering related discussions in https://github.com/pytest-dev/pytest/issues/5743, https://github.com/pytest-dev/pytest-qt/issues/426, and https://github.com/pytest-dev/pytest/issues/10959, I believe this is a problem that should be addressed by `pytest`.

For this project, I suggest adding the `-p no:faulthandler` option.

However, since [`-p no:faulthandler` became a standard feature in `pytest`](https://docs.pytest.org/en/stable/changelog.html#id536) starting from [`pytest==5.0.0`](https://pypi.org/project/pytest/5.0.0/), which [does not support Python 2.7](https://github.com/pytest-dev/pytest/blob/5.0.0/setup.cfg#L25-L29), I believe this kind of branching is necessary.

```suggestion
    if ($env:PYTHON_VERSION -match "2.7") {
        pytest --junit-xml=$results --tb=native --capture=sys --show-capture=no -v --verbosity=3 --cache-clear --durations=15 --ignore=testall.py --log-level=DEBUG --cov-report html:Coverage_report --cov=pywinauto pywinauto\unittests
    } else {
        pytest --junit-xml=$results --tb=native --capture=sys --show-capture=no -v --verbosity=3 --cache-clear --durations=15 --ignore=testall.py --log-level=DEBUG --cov-report html:Coverage_report --cov=pywinauto -p no:faulthandler pywinauto\unittests
    }
```

junkmd (commented):
I want to share some findings from the CI pipeline after it finished, as well as some insights from my own experience.

ci/runTestsuite.ps1
```diff
@@ -1,54 +1,37 @@
...
     cd $env:APPVEYOR_BUILD_FOLDER
-    $stylesheet =  "./ci/transform_xunit_to_appveyor.xsl"
-    $input = "nosetests.xml"
-    $output = "transformed.xml"
-    
-    #nosetests  --all-modules --with-xunit pywinauto/unittests
-    nosetests --nologcapture --exclude=testall --with-xunit --with-coverage --cover-html --cover-html-dir=Coverage_report --cover-package=pywinauto --verbosity=3 pywinauto\unittests
-    $success = $?
-    Write-Host "result code of nosetests:" $success
+    $results = "results.xml"
```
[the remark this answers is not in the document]
vasily-v-ryabov: Thanks! I made it for newest image only to reduce other risks.

ci/runTestsuite.ps1
vasily-v-ryabov: Thanks! It is better to use a variable to be more typo resistant. Done.

vasily-v-ryabov:
0.7.0 will be a major release which includes the whole current branch `atspi` with some additional minor fixes. Big PRs with next features like `backend="wpf"` will be 0.8.0+.

junkmd:
> 0.7.0 will be a major release which includes the whole current branch atspi with some additional minor fixes.

If so, I propose merging `master` into the `atspi` branch after this PR is merged, followed by adding commits to update the CI pipeline and package settings of the `atspi` branch to support only Python 3.7+.

After that, I plan to rebase or resubmit the PRs I’ve submitted so far and allocate my resources to tasks like type annotations and Mac/Linux support.

pywinauto/xml_helpers.py
```diff
@@ -40,7 +40,15 @@
...
 import re
-import bz2, base64
+import bz2
+
+try:
+    from base64 import encodestring
+    from base64 import decodestring
+except ImportError:
+    from base64 import encodebytes as encodestring
+    from base64 import decodebytes as decodestring
```
junkmd, line 50: Since `encodestring` and `decodestring` were deprecated starting with Python 3.9, I believe using a version bridge with `sys.version_info` is better than catching an `ImportError`.

Additionally, considering a future point when only Python 3.9+ will be supported, it would be better to use the existing names instead of continuing to rely on the deprecated ones.

https://docs.python.org/3/whatsnew/3.9.html#removed

```suggestion
import bz2
import sys

if sys.version_info >= (3, 9):
    from base64 import encodebytes
    from base64 import decodebytes
else:
    from base64 import encodestring as encodebytes
    from base64 import decodestring as decodebytes
```

pywinauto/xml_helpers.py
```diff
@@ -100,7 +108,7 @@ def _set_node_props(element, name, value):
                 raise MemoryError
 
             #print('type(value) = ' + str(type(value)))
-            image_data = base64.encodestring(bz2.compress(value.tobytes())).decode('utf-8')
+            image_data = encodestring(bz2.compress(value.tobytes())).decode('utf-8')
```
junkmd, line 111:
```suggestion
            image_data = encodebytes(bz2.compress(value.tobytes())).decode('utf-8')
```

pywinauto/xml_helpers.py
```diff
@@ -408,7 +416,7 @@ def _read_xml_structure(control_element):
 
             # get image Attribs
             img = _get_attributes(elem)
-            data = bz2.decompress(base64.decodestring(img['data'].encode('utf-8')))
+            data = bz2.decompress(decodestring(img['data'].encode('utf-8')))
```
junkmd, line 419:
```suggestion
            data = bz2.decompress(decodebytes(img['data'].encode('utf-8')))
```

junkmd:
I believe [the failure of `DatetimeTestCases.testGetTime`](https://ci.appveyor.com/project/pywinauto/pywinauto/builds/51266205/job/x4y18cemoeg8ndqx#L813) is due to the time when the test was executed, so retrying should resolve the issue.

junkmd:
As confirmed in #1345, Python 3.9 and later are not included in the default AppVeyor build image, and specifying `APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2022` is required.

We need to remember this when modifying the CI pipeline for the `atspi` branch after this PR is merged.

junkmd:
I believe some tests in `test_taskbar.TaskbarTestCases` will pass if we cherry-pick 977f0a0e and 29e4d0e7.

vasily-v-ryabov:
Task bar has significant changes in Win 11 and WinServer 2019+. Currently I skipped some taskbar tests conditionally, so they are covered on older images. 2 TreeView tests are left to fix on image `Visual Studio 2019`.

junkmd:
As seen with [the failure of `testClickCustomizeButton`](https://ci.appveyor.com/project/pywinauto/pywinauto/builds/51267916/job/q7k9bvxu5h0jgbfk/tests) in 5d1a2a0, the tests in `test_taskbar.TaskbarTestCases` have been confirmed to be unstable due to timing issues, as confirmed in #1345 and #1398.

I propose setting the `Timings` to `slow` for all tests in `TaskbarTestCases`.

```diff
--- a/pywinauto/unittests/test_taskbar.py
+++ b/pywinauto/unittests/test_taskbar.py
@@ -173,7 +173,7 @@ class TaskbarTestCases(unittest.TestCase):
 
     def setUp(self):
         """Set some data and ensure the application is in the state we want"""
-        Timings.defaults()
+        Timings.slow()
 
         self.tm = _ready_timeout
         app = Application(backend='win32')
@@ -200,6 +200,8 @@ class TaskbarTestCases(unittest.TestCase):
         except(ProcessNotFoundError):
             l.log("No more leftovers. All good.")
 
+        Timings.defaults()
+
     def testTaskbar(self):
         # just make sure it's found
         taskbar.TaskBar.wait('visible', timeout=self.tm)
```

vasily-v-ryabov:
If the problem is rare enough, I'd prefer to re-start incomplete tasks. Not all such issues are timing related. Ideally it would be nice to upload screenshot at the moment of failure (in `tearDown`?) to better catch a flaky test reason. Feel free to create an issue with label "internal" for such task. `appveyor.yml` contains some code for that at the bottom. Maybe only the first 1 or 2 failures are important enough to take a screenshot because massive failures may lead to a flood of screenshots.

junkmd:
> If the problem is rare enough, I'd prefer to re-start incomplete tasks. Not all such issues are timing related.

Okey, Understood.

Indeed, we often encounter rare issues. In my own projects, I’ve found that adjusting timings can become an endless task. If this issue occurs again, we can manually rerun the job and focus on addressing other problems related to this PR.

Uploading a screenshot as an artifact in case of failure is an excellent idea. I’ll read the AppVeyor documentation.

vasily-v-ryabov:
Woohoo! CI passed! Though one test took ~9 minutes for Python 3.12. :)

```
============================ slowest 15 durations =============================
551.99s call     pywinauto/unittests/test_uiawrapper.py::TreeViewWpfTests::test_tv_drag_n_drop
22.36s call     pywinauto/unittests/test_uiawrapper.py::ListViewWrapperTests::test_cells
```

junkmd:
> merging master into the atspi branch after this PR is merged, followed by adding commits to update the CI pipeline and package settings of the atspi branch to support only Python 3.7+.

I will work for this.

PR #1435, opened 2025-03-22
Cherry picking commits from atspi into master branch

vasily-v-ryabov:
This is another attempt to merge 2 big branches!

 * The main problem is that `atspi` was initially forked from `master` branch at tag `0.6.1` which was significantly improved till tag `0.6.9`.
 * So many commits in `master` and `atspi` are the same but Git or TortoiseGit cannot detect they are really the same. So rebase process is really hard because of that.

Possible solution in this PR:
 * I've forked from recent `master` branch, and I'm cherry-picking only necessary (new) commits from `atspi` branch step by step. When I cherry pick next bunch of commits, I push them to this PR and make sure it is still mergeable. So I can detect where the mergeability problem starts using such approach. Or maybe I'm lucky. 😄  Fingers crossed. 🤞


vasily-v-ryabov:
Ha-ha-ha! Exactly 100 commits can be merged by rebase method in a PR. I'm not sure this is a bug or intended limitation. So my plan is merging step by step 100 commits each time. This is a little bit risky for `master` branch. So I will create `master_backup` branch. But risk of all tags corruption still remains... Hmm...

PR #1436, opened 2025-03-22
Atspi rebase on master  faster image

vasily-v-ryabov:
Somehow AppVeyor image `Visual Studio 2015` runs pywinauto tests faster and more reliably. This version of Windows Server provides legacy version of Notepad, SysTray and other things which were updated in newer Windows images. So moving tests to newer Windows apps or to separate self-built apps could be a separate task now.

This PR is huge for review, so if someone can take a look at last 2-3 commits, it would be good enough. All previous commits to `atspi` branch were already reviewed previously, so they were all rebased with some bunch of issues which I fixed by separate commits.


vasily-v-ryabov:
It seems mergeable in fact, but I'm planning to push these changes directly to `master` branch after Windows CI has passed (tomorrow). It should automatically mark this PR as merged.

PR #1438, opened 2025-03-30
Make linux tests stable.

junkmd:
When I rebased my past PR, the CI pipeline that ran failed the tests on Linux, even though the changes did not modify the atspi functionality or cause any side effects.

After investigating the cause, it seems that an exception was likely being thrown in `AtspiElementInfoTests.setUp`.
To improve stability, I extended the sleep duration.

Once this is merged, I plan to rebase my past PR onto it as well.

### DIFF
--- a/.github/workflows/atspi_test.yml
+++ b/.github/workflows/atspi_test.yml
@@ -86,6 +86,6 @@ jobs:
         sleep 5
     - name: Run tests
       run: |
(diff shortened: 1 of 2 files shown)


vasily-v-ryabov (approved):
Thanks! This is fine for now. Though we will need some wait for application window in `atspi` backend during `Application(...).start(...)` like it was done a long time ago for Windows `Application`.

PR #1440, opened 2025-04-04
Add and fix UIA pattern names.

junkmd:
I noticed that in the UIA pattern names, `IUIAutomationDropPattern` is absent, and [`IUIAutomationDropTargetPattern`](https://learn.microsoft.com/en-us/windows/win32/api/uiautomationclient/nn-uiautomationclient-iuiautomationdroptargetpattern) exists instead. I also noticed that [`IUIAutomationSelectionPattern2`](https://learn.microsoft.com/en-us/windows/win32/api/uiautomationclient/nn-uiautomationclient-iuiautomationselectionpattern2) is not covered.

I vaguely remembered committing this fix and addition to my local branch a long time ago.

I've rebased those changes from that time onto the current `master`.

### DIFF
--- a/pywinauto/windows/uia_defines.py
+++ b/pywinauto/windows/uia_defines.py
@@ -134,7 +134,7 @@ def _build_pattern_ids_dic():
         'Transform',
 
         # Windows 8 and later
-        'Annotation', 'Drag', 'Drop', 'ObjectModel', 'Spreadsheet',
+        'Annotation', 'Drag', 'DropTarget', 'ObjectModel', 'Spreadsheet',
         'SpreadsheetItem', 'Styles', 'TextChild', 'TextV2', 'TransformV2',
 
         # Windows 8.1 and later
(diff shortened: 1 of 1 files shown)


vasily-v-ryabov (commented):
We should be careful about new pattern names since they may be absent in older versions of `UIAutomationCore.dll`. I remember it was some condition for such cases to avoid exception at UIA backend initialization. But I don't recall where it was exactly. :)

vasily-v-ryabov (approved):
Well, it seems the correct place. `hasattr` check is already here. Nevermind.

vasily-v-ryabov:
BTW, the full list of patterns is probably here: https://learn.microsoft.com/en-us/windows/win32/winauto/uiauto-client-controlpatterninterfaces
On each page there is a detailed info about supported desktop and server OS versions. It may be useful for the future re-review.

PR #1461, opened 2025-08-22
exit() being used instead of sys.exit

Adamdoesdumbcode:
prettu self explanitory. in wireshark.py "exit()" was used instead of sys.exit. the entire file had sys.exit except this one part so i was like might as well


vasily-v-ryabov:
Thanks! Finally I decided to merge https://github.com/pywinauto/pywinauto/pull/1462 since extra empty line is not needed. Closing this PR. Don't worry and welcome for other contributions.

PR #1468, opened 2025-09-30
Improved capture_as_image with DWM bounds and cropping

Jlee547:
In this PR, I refined the capture_as_method function in win_base_wrapper.py to improve the screenshot accuracy for windowed applications. It uses DwmGetWindowAttribute with DWMWA_EXTENDED_FRAME_BOUNDS to cpature the full window, and has a manual pixel trimming to eliminate border artifacts. 
This is the before:
notepad_screenshot

This is the after:
edge_dwm_bounds


Jlee547:
Hi, I updated the script and the latest output looks like this:
edge_dwm_bounds

The tighter pixel trims introduced several errors, mostly that the borders were not as expected. However, the trimming logic is working as intended, and the visual result is consistent.

PR #1477, opened 2026-01-10
clarify clipboard limitations in documentation

Darshangouda55:


vasily-v-ryabov:
Thanks for the contribution attempt. I think the bottom of Readme.md is improper place for such comment. It would be better to add such comment to `clipboard` or `send_keys` module documentation.

PR #1482, opened 2026-03-30
Fix/dependabot security alerts

jasonbogdanskiACN:
## Summary

Resolves all 43 open Dependabot security alerts by updating vulnerable dependency version constraints in \dev-requirements.txt\.

### Pillow
- **Python <=3.7**: \6.2.0\ → \9.5.0\ (last version to support Python 3.7)
  - Fixes ~20 high/critical CVEs across Pillow 6.x–9.x
- **Python ==3.8**: loosened pin from \==10.3.0\ to \>=10.3.0\ (Pillow 11+ dropped Python 3.8; 10.3.0 remains the highest safe version)
  - No functional change, just ensures pip resolves to a non-vulnerable release
- **Python >3.8**: already at \12.1.1\ — no change needed

### pywin32 (alert #16)
- Python <=3.6 line: changed from \<=227\ (vulnerable) to \>=301\
- Python >3.6 line: bumped minimum from \>=300\ to \>=301\
- Fixes integer overflow CVE resolved in pywin32 301

## Testing
No code changes — only version constraints updated. CI (AppVeyor) will validate across all supported Python versions.

PR #1483, opened 2026-04-11
ci: pin codeql workflow actions to immutable SHAs

grtninja:
## Summary
- pin the four floating actions in `.github/workflows/codeql.yml` to immutable SHAs
- preserve the existing CodeQL trigger matrix and job behavior
- keep this as a one-file workflow hardening change only

## Validation
- `git diff --check`
- `python -c "import pathlib, yaml; p=pathlib.Path(r'.github/workflows/codeql.yml'); yaml.safe_load(p.read_text(encoding='utf-8')); print('yaml_ok', p)"`
- final exact-file overlap recheck found no open PR already touching `.github/workflows/codeql.yml`

## Notes
- no repo-local CONTRIBUTING/SECURITY/CODEOWNERS guidance was present on the working branch, so this stayed strictly one-file and behavior-preserving
- this does not overlap the adjacent `security_audit.yml` / `SECURITY.md` work in #1482
- prepared with AI assistance and manually validated before submission

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -24,18 +24,18 @@ jobs:
 
     steps:
       - name: Checkout
-        uses: actions/checkout@v3
+        uses: actions/checkout@f43a0e5ff2bd294095638e18286ca9a3d1956744
 
       - name: Initialize CodeQL
-        uses: github/codeql-action/init@v2
+        uses: github/codeql-action/init@b8d3b6e8af63cde30bdc382c0bc28114f4346c88
         with:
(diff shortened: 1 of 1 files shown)


vasily-v-ryabov:
I don't understand the need of such change. And it even looks suspicious to me since GitHub Actions can't automatically update workflows if they found it vulnerable one day. So I'm strictly against such pinning. Yes, there is no guarantee that fresh changes are not vulnerable to new PRs. But I don't have time resources to keep these SHAs up-to-date.

PR #1484, opened 2026-04-11
fix: accept WindowSpecification in find_elements parent

### LINKED ISSUE #1313: Cannot pass WindowSpecification to find_elements(parent=...)
## Expected Behavior
find_elements should return descendants of a window with a parent argument.

## Actual Behavior
I was trying to pass a `WindowSpecification` instance to the parent argument of find_elements, but find_elements raised exceptions. 

## Steps to Reproduce the Problem
1. In this code snippet, `elements` is always empty. 

```python
app = Application(backend="uia").connect(process=1964)
window = app.top_window()
elements = pywinauto.findwindows.find_elements(
    backend="uia",
    process=app.process,
    parent=window,
    top_level_only=False,
)
```

2. If I remove the process argument, in this code snippet, an exception would be raised. 

```python
app = Application(backend="uia").connect(process=1964)
window = app.top_window()
elements = pywinauto.findwindows.find_elements(
    backend="uia",
    parent=window,
    top_level_only=False,
)
```

The traceback is attached below: 
```
Exception has occurred: AttributeError       (note: full exception trace is shown but execution is paused at: _run_module_as_main)
'UIAWrapper' object has no attribute 'visible'
  File "D:\Workspace\XXX\.venv\Lib\site-packages\pywinauto\findwindows.py", line 286, in <listcomp>
    elements = [elem for elem in elements if elem.visible]
```

## Short Example of Code to Demonstrate the Problem
I have investigated this issue and I found that the parent argument must be passed in a way showed below: 

```python
elements = pywinauto.findwindows.find_elements(
    backend="uia",
    process=app.process,
    parent=window.element_info,
    top_level_only=False,
)
```

I am wondering if this is an expected behavior or a bug. If this is intended, I would like to suggest that to document it in the docstring.  

# Specification 
  - Pywinauto version: 0.6.8
  - Python version and bitness: 3.11.2
  - Platform and OS: Windows 11 x64

grtninja:
## Summary
- accept `WindowSpecification` instances for `find_elements(parent=...)`
- resolve the parent through `parent.find().element_info` before the existing parent type gate
- add regression coverage showing `parent=<WindowSpecification>` matches the established `parent=<element_info>` behavior for both `win32` and `uia`

## Why
Fixes #1313. Higher-level search paths already understand `WindowSpecification`, so rejecting it in `find_elements(parent=...)` is inconsistent and forces callers onto a lower-level workaround.

## Validation
- `python -m unittest pywinauto.unittests.test_findwindows`

## Risk Notes
- narrow parent-resolution change only
- no touch to `base_application.py`
- no broader search-behavior refactor

### DIFF
--- a/pywinauto/findwindows.py
+++ b/pywinauto/findwindows.py
@@ -140,6 +140,16 @@ def find_elements(**kwargs):
         backend = registry.active_backend.name
     backend_obj = registry.backends[backend]
 
+    # Allow passing WindowSpecification as a parent (issue #1313).
+    # Resolve it early to keep the rest of this function working with element_info.
+    if parent is not None:
+        try:
+            from .base_application import WindowSpecification
+        except Exception:  # pragma: no cover
+            WindowSpecification = ()
+        if isinstance(parent, WindowSpecification):
+            parent = parent.find().element_info
+
     if handle is not None:
         # TODO: uncomment later
         #if not kwargs:
--- a/pywinauto/unittests/test_findwindows.py
+++ b/pywinauto/unittests/test_findwindows.py
@@ -36,7 +36,6 @@ import unittest
 
 import sys, os
 sys.path.append(".")
-from pywinauto.backend import registry
 from pywinauto.windows.application import Application
 from pywinauto.sysinfo import is_x64_Python
 from pywinauto.findwindows import find_window, find_windows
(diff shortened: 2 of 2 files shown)


vasily-v-ryabov:
First of all, generally I encourage to use only high level search because `find_elements` needs deep knowledge of how to select specific backend which is for very advanced QA automation guys (not typical in 99% cases). And such low level programming doesn't add real new possibilities, only pain for newcomers. I personally don't ever use `find_elements` directly when I use pywinauto.

Another thing is a proper dependencies structure inside the library which is broken by circle import introduced by this PR. Yes, current structure is also not ideal and it has some circle imports too. But I'd prefer keeping its number not growing. So my main concern is "find_elements shouldn't know about higher level WindowsSpecification because it is mixing architecture layers".

Thank you for understanding!

